Issue #295, opened 2017-02-15
Attribute Editor too narrow


metas-mk:
### Is this a bug or feature request?
Bug.
### What is the current behavior?
The Attribute Editor in Material Receipt Workflow is too narrow. The Field names are not recognizable anymore.
#### Which are the steps to reproduce?
Open the Handling Unit editor in Window Material Receipt Candidates.
### What is the expected or desired behavior?
Have a minimum width or make responsive (label above the field when too narrow screen) for the Attribute editor to be able to recognize the field names.
![image](https://cloud.githubusercontent.com/assets/13365687/22962576/f528a450-f34d-11e6-96f3-0f9b4c1f3a10.png)

metas-mk:
Test. Works nicely now. Attribute widget is now wide enough (could be even narrower now).

metas-mk:
Added to upcoming Release 4.56.55

PR #305, opened 2017-02-15
#295 Fix table-content-empty css class toggling

damianprzygodzki:


src/components/table/Table.js
```diff
@@ -668,12 +668,12 @@ class Table extends Component {
 
                     <div
                         className={
-                            "panel panel-primary panel-bordered panel-bordered-force table-flex-wrapper document-list-table" +
+                            "panel panel-primary panel-bordered panel-bordered-force table-flex-wrapper document-list-table " +
                             ((
                                 (rowData && rowData[tabid] &&
                                 Object.keys(rowData[tabid]).length === 0) ||
                                 (!rowData[tabid])
-                            ) ? " table-content-empty" : null)
```
damianprzygodzki: @chmieludaniel If you are toggling css class:
- please, keep consistent the project schema, always the one free space char is *after* toggled class
- return empty string instead of null

PR #308, opened 2017-02-15
Add fork shortcuts library, fix blur lookup #214

Dunkat:
'npm install' - to install new shortcuts library

### DIFF
--- a/src/components/widget/Lookup/Lookup.js
+++ b/src/components/widget/Lookup/Lookup.js
@@ -203,11 +203,14 @@ class Lookup extends Component {
         dispatch(openModal("Add new", windowType, "window"));
     }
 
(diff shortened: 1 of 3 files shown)


src/components/widget/Lookup/Lookup.js
```diff
@@ -203,16 +203,11 @@ class Lookup extends Component {
         dispatch(openModal("Add new", windowType, "window"));
     }
 
-    handleBlur = (callback) => {
+    handleBlur = () => {
 
         this.setState(Object.assign({}, this.state, {
             isOpen: false
-        }), () => {
```
damianprzygodzki: We need that for toggle `this.clearState`

PR #328, opened 2017-02-16
Quick action button responsive #315

Dunkat:


src/assets/css/variables.css
```diff
@@ -34,4 +34,5 @@ $checkbox-unchecked: rgba(107, 123, 134, 0.2);
 
 $tooltip-border-color: #d5e5f0;
 
+$breakpoint-md: 840px;
```
damianprzygodzki: Shouldn't be here some existing breakpoint like 991px ? Or level down 768px?

Dunkat: changed to 991

Issue #161, opened 2016-12-20
Combined Fields behaviour enhancement


metas-mk:
Combined Fields behaviour enhancement
- Example BPartner Field in Sales Order. Is a combined Field with Businesspartner, Location and Contact. When selecting the Business Partner, then Location and Contact are set initially, so in those cases the user could somehow "jump" out of that field without further recording. Please enable that with <TAB>

![image](https://cloud.githubusercontent.com/assets/13365687/21352832/9f471320-c6c3-11e6-909f-0e7e2c5ad892.png)

cadavre:
1) After filling BPartner, if rest of fields were filled automatically, dropdown shall close and let user [Tab] to next input below.
2) After filling BPartner, if only Location was filled, dropdown shall jump automatically to Contact person and after selection let user [Tab] to next input.
3) After filling BPartner, if only Contact was filled, dropdown shall jump automatically to Location and after Location selection dropdown shall close and let user [Tab] to next input.

4) If focus is on Combined fields input but no dropdown is shown – arrows up/down shall open dropdown, arrows left/right shall switch between fields.

Dunkat:
Could you provide business partners which has no automatically filled all fields? because I can't find any.

metasnw:
1. if the address is missing entirely there will be no autofill but an error message
2. if there is an address but flag shipment address is NO there will be no autofill but you also can not manually select the address

you can try partner "noautofil" which behaves like No. 2

Dunkat:
Ok thanks, using "noautofil" generete error "Business Partner has no Bill To Address". So there is two options: autofill or no possible to select manually fields, right?

metas-lc:
![autofill](https://cloud.githubusercontent.com/assets/16001666/23158722/e2ef0286-f828-11e6-968f-c2d4467f692c.png)

IT (questions)
1. i made a sales order with this bpartner that has no address (at all)
=> i got the error "Business Partner has no Bill To Address - Business Partner: 1000010_noautofil (ID=2156498)"
2. i added an address to the bpartner, but is not shipment/invoice address. (no checkbox set)
- i got the address set in invoice partner and business partner. Is that wanted? 
(in business partner the address appeared only after i refreshed the page)
- shouldn't i get the same message "Business Partner has no Bill To Address"?

damianprzygodzki:
[...]

PR #336, opened 2017-02-17
Combined Fields behaviour enhancement #161

Dunkat:


src/components/widget/Lookup/Lookup.js
```diff
@@ -160,10 +161,18 @@ class Lookup extends Component {
...
     }
 
+
     generatingPropsSelection = () => {
         const {onChange} = this.props;
         const {properts} = this.state;
         const propertiesKeys = Object.keys(properts);
+        let isClose = true;
+
+        this.props.defaultValue.map (item => {
```
damianprzygodzki: If we are using props, than escape them to a new const.

src/components/widget/Lookup/Lookup.js
```diff
@@ -160,10 +161,18 @@ class Lookup extends Component {
...
+            if(!item.value){
```
damianprzygodzki: this whole mapping, isn't it just boolean value of length of filtered array with !item.value?

src/components/widget/Lookup/LookupList.js
```diff
@@ -53,6 +53,7 @@ class LookupList extends Component {
         const {handleSelect, selected} = this.props;
         const name = item[Object.keys(item)[0]];
         const key = Object.keys(item)[0];
+        // console.log(item);
```
damianprzygodzki: No console logs in PRs! Even commented.

PR #360, opened 2017-02-20
Strange pulse effect fix #350

### LINKED ISSUE #350: Strange pulse effect in Subtab Gridview
Bug.
When selecting one row in a subtab going back to the main view, the wrong line is pulsed.
Go to Business Partner Window. Subtab Location w/ multiple rows. Select the first line. Advanced edit. Dont't edit anything. Done. The last row is pulsed in the main view.
When pulsing, then pulse the correct line. When not changing anything, then in my pov no line shall be pulsed.
image
image

Dunkat:


src/containers/MasterWindow.js
```diff
@@ -22,7 +22,7 @@ class MasterWindow extends Component {
         }
     }
 
-    closeModalCallback = (isNew) => {
+    closeModalCallback = (type, isNew) => {
```
damianprzygodzki: I am bit confused, why we are declaring variable that is never used?

So i checked, and i see that closeCallback for modal has a lot useless parameters, shouldn't we get rid of them instead?

src/containers/MasterWindow.js
```diff
@@ -112,7 +112,7 @@ class MasterWindow extends Component {
                         isAdvanced={modal.isAdvanced}
                         viewId={null}
                         closeCallback={(isNew) => this.closeModalCallback(
```
damianprzygodzki: And in that case, what is the reason to wrap method execution in another function and passing the same, one and only parameter?

Issue #382, opened 2017-02-21
Arrow key not working in dropdown


metas-mk:
### Is this a bug or feature request?
Bug
### What is the current behavior?
When tabbing to a field w/ is the dropdown, the selection via arrow down/ up is not possible anymore.
#### Which are the steps to reproduce?
Try window Purchase Order. Tab to Price System, try arrow.
### What is the expected or desired behavior?
Make the arrows usable again.
![image](https://cloud.githubusercontent.com/assets/13365687/23160822/36acfa82-f828-11e6-8c08-69639473c9c9.png)

metas-lc:
Created new task for the scrollbar. Moving this task to done
https://github.com/metasfresh/metasfresh-webui-frontend/issues/416

metas-lc:
docu https://github.com/metasfresh/metasfresh-documentation/blob/master/_tests_collection/Testcase_%23382.md

PR #390, opened 2017-02-21
Fix dropdown navigation #382

Dunkat:


src/components/widget/List/RawList.js
```diff
@@ -60,18 +79,24 @@ class RawList extends Component {
...
-        const {selected} = this.state;
-        const {list} = this.props;
+        const {selected, dropdownList} = this.state;
+
+        let selectedIndex = null;
 
-        const next = up ? selected + 1 : selected - 1;
+        dropdownList.map((item, index) => {
+            if(JSON.stringify(item) === JSON.stringify(selected)){
+                return selectedIndex = index;
```
damianprzygodzki: What you wanted to do here? It looks no good.

src/components/widget/List/RawList.js
```diff
@@ -8,7 +8,26 @@ class RawList extends Component {
...
+
+        if(prevProps.list !== list){
+            let dropdown = [];
+
+            if(!mandatory){
+                dropdown.push(0);
+            }
+
+            if(list.length > 0) {
+                this.setState(Object.assign({}, this.state, {
```
damianprzygodzki: @Dunkat We are going away from the assumption, to always provide deep copy of object, event if it is one level deep. `setState` is doing that for us.

src/components/widget/List/RawList.js
Dunkat: comparing two objects

Dunkat: and return index of item which is selected to  have possibility get higher or lower index

damianprzygodzki: So i believe the assignment `return selectedIndex = index;` is redundant.

src/components/widget/List/RawList.js
```diff
@@ -60,18 +79,25 @@ class RawList extends Component {
...
-            selected: (next >= 0 && next <= list.length) ? next : selected
+        dropdownList.map((item, index) => {
+            if(JSON.stringify(item) === JSON.stringify(selected)){
+                selectedIndex = index;
+            }
         });
+
+        const next = up ? selectedIndex + 1 : selectedIndex - 1;
+
+        this.setState(Object.assign({}, this.state, {
```
damianprzygodzki: The same thing, Object.assign to remove

PR #414, opened 2017-02-23
Fix wrong viewId in material receipt #404

Dunkat:


src/components/app/QuickActions.js
```diff
@@ -125,7 +127,7 @@ class QuickActions extends Component {
                         }
                     </div>
                     <QuickActionsContextShortcuts
-                        handleClick={() => this.handleClick(actions[0])}
+                        handleClick={() => shouldNotUpdate ? '' : this.handleClick(actions[0])}
```
damianprzygodzki: Null will be much better here then empty string

damianprzygodzki: Or just make a conjunction

PR #417, opened 2017-02-23
Fix depth param and child param in navigation tree #375

Dunkat:


src/containers/NavigationTree.js
```diff
@@ -39,7 +39,7 @@ class NavigationTree extends Component {
 
     getData = (callback) => {
         const {dispatch} = this.props;
-        dispatch(rootRequest()).then(response => {
+        dispatch(rootRequest('', '0')).then(response => {
```
damianprzygodzki: First, why it is a string?

src/actions/MenuActions.js
```diff
@@ -72,10 +72,11 @@ export function queryPathsRequest(query, limit, child) {
     );
 }
 
-export function rootRequest(limit) {
+export function rootRequest(limit, depth) {
     return () => axios.get(
         config.API_URL +
-        '/menu/root?depth=10' +
+        '/menu/root?depth=' +
+        (depth ? depth : '10')  +
```
damianprzygodzki: Why the default value is set to 10? Shouldnt we fix 0 as a default value? To understand that like, `always fetch all until you pass something` And if we want to fix some value, shouldn't be better option to use defaults for parameters in functions like:
```
export function rootRequest(limit, depth = 0) {
...
'/menu/root?depth=' +
depth
```

Because rootRequest is (for now) used in one special case (that is using default value). So we are going to save passing some parameters below :)

src/containers/NavigationTree.js
Dunkat: now string is not needed  :)

src/actions/MenuActions.js
Dunkat: fixed

PR #434, opened 2017-02-24
fix #416 - Scroll inside dropdown list on arrow key navigation

### LINKED ISSUE #416: moving scrollbar on arrow key in dropdown
### Type of issue

Bug

### Current behavior
use arrow key in a dropdown that contains lots of data. The scrollbar dows not move (up/down) so i cannot see what i'm selecting
### Expected behavior
scrollbar shall follow my selection in dropdown
### Steps to reproduce
1. go to Product window and select a product (single view)
2. select Statistics Group field 
3. move in dropdown with arrow keys => the scollbad does not move to allow me to see the selection

**Window:** enter window ID or Navigation menu step-by-step.
140
**Entity:** enter document ID or steps to reproduce environment.
1000001

**Link:** 
http://w101.metasfresh.com:8090/window/140/1000001

chmieludaniel:


src/components/widget/List/RawList.js
```diff
@@ -29,6 +29,40 @@ class RawList extends Component {
...
         }
+
+        const { isOpen, selected } = this.state;
+        if (!isOpen) {
+            return;
+        }
+
+        const {listScrollWrap, items} = this.refs;
+
+        const listElementHeight = 30;
```
damianprzygodzki: We should at least get this value from element - it can change.

PR #530, opened 2017-03-13
Major Sidelist error #528

Dunkat:
note: hotfix - prevent of passing undefined windowType (master.layout.type in MasterWindow.js)

### DIFF
--- a/src/components/app/DocumentList.js
+++ b/src/components/app/DocumentList.js
@@ -195,20 +195,22 @@ class DocumentList extends Component {
             viewId
         } = this.state;
 
-        dispatch(initLayout(
-            'documentView', windowType, null, null, null, null, type, true
-        )).then(response => {
-            this.setState({
-                layout: response.data
-            }, () => {
-                if(viewId && !isNewFilter){
-                    this.browseView();
-                }else{
-                    this.createView();
-                }
-                setModalTitle && setModalTitle(response.data.caption)
-            })
-        });
+        if(windowType){
+            dispatch(initLayout(
+                'documentView', windowType, null, null, null, null, type, true
+            )).then(response => {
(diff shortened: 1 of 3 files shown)


src/components/app/DocumentList.js
```diff
@@ -195,20 +195,22 @@ class DocumentList extends Component {
...
-            }, () => {
-                if(viewId && !isNewFilter){
-                    this.browseView();
-                }else{
-                    this.createView();
-                }
-                setModalTitle && setModalTitle(response.data.caption)
-            })
-        });
+        if(windowType){
```
damianprzygodzki: Is it needed for sure? Because in PropTypes we have defined `isRequired` type. So there should be no valid component until windowType prop delivered.

src/components/app/DocumentList.js
```diff
@@ -379,7 +381,7 @@ class DocumentList extends Component {
                             readonly={true}
                             keyProperty="id"
                             onDoubleClick={(id) => {
-                                !isModal &&
+                                !isModal && windowType &&
```
damianprzygodzki: The same here.

src/components/header/Header.js
```diff
@@ -396,7 +396,7 @@ class Header extends Component {
                 />}
 
                 {showSidelist && isSideListShow && <SideList
-                    windowType={parseInt(windowType)}
+                    windowType={windowType ? parseInt(windowType) : ''}
```
damianprzygodzki: And here.

src/components/app/DocumentList.js
Dunkat: changed solution, the problem was because SideList was opened after changing document view, and windowType prop was undefined

PR #562, opened 2017-03-22
Closing modal or window with unsaved changes #453

Dunkat:


src/components/app/Modal.js
```diff
@@ -117,6 +117,22 @@ class Modal extends Component {
...
+
+        if(!modalSaveStatus){
+            const result = window.confirm('Do you really want to leave?');
+
+            if(result){
+               this.closeModal();
+            }
+        } else {
+            this.closeModal();
+        }
```
damianprzygodzki:
```
if(modalSaveStatus || window.confirm('Do you really want to leave?')){
      this.closeModal();
}
```

What about this :) Bit shorter and less statements.

Dunkat: good point, fixed

PR #566, opened 2017-03-22
#541 includedTabsInfo support

damianprzygodzki:


damianprzygodzki:
Ready to merge, but waiting for tested #596 and it should be merged after.

PR #663, opened 2017-04-20
Handle process layout type field #655

Dunkat:


src/components/app/OverlayField.js
```diff
@@ -0,0 +1,62 @@
+import React, { Component } from 'react';
+import MasterWidget from '../widget/MasterWidget';
+
+import {
+    findRowByPropName
+} from '../../actions/WindowActions';
+
+class OverlayField extends Component {
+
```
damianprzygodzki: Oh my god ;p How it is working without actually passing props to constructor of parent class? super(props) is missing!

PR #674, opened 2017-04-24
Process error not displayed #672

Dunkat:


src/containers/App.js
```diff
@@ -61,10 +61,8 @@ axios.interceptors.response.use(function (response) {
 
             const errorTitle = errorMessenger(status);
 
-            console.error(data.message);
```
damianprzygodzki: Please, revert this.

PR #728, opened 2017-05-11
Implement Zoom-into in webUI #720

Dunkat:


src/components/widget/RawWidget.js
```diff
@@ -747,4 +755,10 @@ class RawWidget extends Component {
     }
 }
 
-export default RawWidget;
+RawWidget.propTypes = {
```
damianprzygodzki: To remove.

PR #791, opened 2017-05-25
Included tab shall preserve the retrieve order #714

Dunkat:


src/actions/WindowActions.js
```diff
@@ -336,12 +336,7 @@ export function getTab(tabId, windowType, docId) {
         dispatch(getData('window', windowType, docId, tabId))
             .then(res => {
                 if(res.data){
-                    let tab = {};
-                    res.data.map(row => {
-                        row.fields = parseToDisplay(row.fields);
```
damianprzygodzki: We have to parse that anyway

PR #794, opened 2017-05-25
Included tab shall preserve the retrieve order - fix PR #714

Dunkat:


src/actions/WindowActions.js
```diff
@@ -336,7 +336,12 @@ export function getTab(tabId, windowType, docId) {
         dispatch(getData('window', windowType, docId, tabId))
             .then(res => {
                 if(res.data){
-                    return res.data;
+                    let tab = [];
```
damianprzygodzki: It can be better, the tab variable is redundant:
```(...).then(res => res.data && res.data.map(row => Object.assign({}, row, {fields: parseToDisplay(row.fields)}))```

File: src/actions/WindowActions.js [hunk 2]
```diff
@@ -335,9 +335,12 @@ export function getTab(tabId, windowType, docId) {
...
         dispatch(getData('window', windowType, docId, tabId))
             .then(res => {
-                if(res.data){
-                    return res.data;
-                }
+                res.data && res.data.map(row => Object.assign({}, row,
+                    {fields: parseToDisplay(row.fields)}
+                ))
+
+                return res.data;
```
damianprzygodzki: this return is also redundant, just do `res => res.data && ...`

PR #805, opened 2017-05-29
Included row: show row's references in context menu #801

Dunkat:


src/components/table/TableContextMenu.js
```diff
@@ -45,14 +55,40 @@ class TableContextMenu extends Component {
         });
     }
 
+    getReferences = () => {
+        const {dispatch, docId, tabId, type, selected} = this.props;
+
+        dispatch(
```
damianprzygodzki: Request without dispatch please.

Issue #885, opened 2017-06-13
Change the Bookmark Subtab Group Functionality


metas-mk:
### Type of issue
Feature request
### Current behavior
Currently, there is this group Link Functionality in Navigation Menu which allows opening the next navigation page to see additional entries. Now that we are using Bookmarks it feels counter intuitive having to click to open the group and then only see the bookmarked entry.
### Expected behavior
Don't have the Subgroups as a link. Show the whole Bookmarked Hierarchy on the first Navigation Menu screen. The Look&Feel shall be exactly like in Sitemap just compressed to the bookmarked items. Here is an example how the Navigation menu should look like then:
![image](https://user-images.githubusercontent.com/13365687/27098937-785f096c-5079-11e7-9f3e-7ffc62345c2f.png)

This Feature Request should be solved together w/ this one: https://github.com/metasfresh/metasfresh-webui-frontend/issues/884

![image](https://user-images.githubusercontent.com/13365687/27091135-8e292918-505f-11e7-8e11-44fc864d81c8.png)

metas-lc:
found this while testing another task
sales order, select an entry -> select breadcrumb
=> you can see the whole menu. NOK
e.g. https://w101.metasfresh.com:8443/window/143/1005597
happens everywhere
![ummm](https://user-images.githubusercontent.com/16001666/27136091-04938c6a-5123-11e7-8c7e-6ac57e3d78cb.png)

damianprzygodzki:
What is the expected behavior here? Because i think, when you have no bookmarks we should display everything from `/api/menu/root?depth=10&childrenLimit=6&favorites=true`. So i think or it is a backend issue, or it is not IT-failed - because that is what we wanted.

metas-mk:
Tested. Works as required for this iteration. 
Follow-up improvement task created for later:
https://github.com/metasfresh/metasfresh-webui-frontend/issues/920

Closing/ Done.

PR #901, opened 2017-06-16
Not render whole menu in breadcrumb children #885

Dunkat:


damianprzygodzki:
There are some style dependencies bugs. More info on tomorrow call.

PR #986, opened 2017-07-05
#971 Page Up and Page Down load all data even if there is only one page

### LINKED ISSUE #971: Page Up and Page Down load all data even if there is only one page
### Type of issue
* Bug
### Current behavior
* If there ar elines in a window that are all included in 1 page, ,pressing the PgUp and PgDn keys load all the data nevertheless
### Expected behavior
* In such case, the data shall not be reloaded 
### Steps to reproduce
* Go to a window with few lines. I tried with Lieferung but it happend everywhere
* GO to a line then start pressing PgUp and PgDn
* => NOK: On each key pressing the data is loaded

image

robertup:
https://github.com/metasfresh/metasfresh-webui-frontend/issues/971
So I've simply made handleNextPage() or handlePrevPage() is not called when pages number is not greater than 1.

Steps to test
Go to any window with few lines (one page) then start pressing PgUp and PgDn
On each key pressing the data is not reloaded

### DIFF
--- a/src/components/shortcuts/PaginationContextShortcuts.js
+++ b/src/components/shortcuts/PaginationContextShortcuts.js
@@ -8,7 +8,7 @@ class PaginationContextShortcuts extends Component {
 
     handleShortcuts = (action, event) => {
         const {
-            handleFirstPage, handleLastPage, handlePrevPage, handleNextPage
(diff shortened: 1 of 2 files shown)


metas-ts:
Thank you @robertup . 
We rolled out the PR and am going to test it today

PR #988, opened 2017-07-06
Fix wrong scope of focusing active table row #968

wiadev:
#968


metas-ts:
rolling it out to the IT server..

PR #1055, opened 2017-07-19
Pick action doesn't properly open included view #1046

wiadev:
#1046

### DIFF
--- a/src/components/app/DocumentList.js
+++ b/src/components/app/DocumentList.js
@@ -186,7 +186,9 @@ class DocumentList extends Component {
             this.setState({
                 cachedSelection: null
             }, () => {
-                dispatch(setListIncludedView());
+                if (includedView.windowType === 'pickingSlot') {
+                    dispatch(setListIncludedView());
+                }
             })
         }
     }


src/components/app/DocumentList.js
```diff
@@ -186,7 +186,9 @@ class DocumentList extends Component {
             this.setState({
                 cachedSelection: null
             }, () => {
-                dispatch(setListIncludedView());
+                if (includedView.windowType === 'pickingSlot') {
```
teosarca: @wiadev  hmmm... not sure this one because 'pickingSlot' it's just one of the many windowTypes.
IMHO it shall work no matter which is the windowType.
If the process/action returned that "an included view shall be opened" then it shall be opened no matter what.

Btw, might be that i am not understanding the underlying problem from here. Pls feel free to discuss about that on chat or even voice call (preferable for me because it's faster).

wiadev: @teosarca 
Yes, this piece of code looks very weird.

Please let me explain.

1. Before my changes this line was intended to reload includedView contents when user clicks on another product line (on left side).
2. But when we added overriding includedView upon click on "Select HU" action we've get triggering of this method call when we actually just replacing the includedView with a new one.
3. So your assumption - "an included view shall be opened" then it shall be opened no matter what" is valid and it will work in the same way. The only exception - pickingSlot window, which has a bit different data flow and needs to clear contents (see line 189 before my changes) and then load new list of slots.

In fact this is workaround to avoid more massive changes in DocumentList.

teosarca: @wiadev 

about 3. I am a bit concerned about that. While the fix is solving our current problem (which is in Picking modal popup) it's not solving it generally. And that's also important for us because these features are basically our building blocks for more and more functionalities (like LEGO :) ).

I think we shall create a new task to refactor and do it right. On that task we shall also discuss about it, i can also give you more input. We don't have any pressure on that. The "high prio" of this task was exactly because we need to deliver the Picking window ASAP.
wdyt?

teosarca: but before i could integrate this task, pls check and fix following issue:

image

i.e. the quickActions endpoint is called using the previous windowId/viewId (i.e. the pickingSlot) instead of the current windowId/viewId (the HU).

wiadev: I told that this is workaround. But doing this right requires significant rework of DocumentList and potentially has wide impact, so we need to do this carefully.

Distinct task for that would be good.

teosarca: for the records: this PR was closed because the fix was provided on https://github.com/metasfresh/metasfresh-webui-frontend/pull/1069

PR #1085, opened 2017-07-26
Fixes incorrect behavior for includedViews and quickActions #992 #1046

wiadev:
#992 #1046


wiadev:
@teosarca check this PR, please.

teosarca:
Checked #1046, but after i call the "Select HU to pick" the left side view (the included one) is empty.
Do you also get this on your workstation?

image

PR #1087, opened 2017-07-26
fix(Filter): keep track of active filters

vko-online:
Problem: filters stored as objects, they override each other every time 1 filter changes
Solution: Keep track of every filter separate, using single uniq key identifier

task: https://github.com/metasfresh/metasfresh-webui-frontend/issues/1021

### DIFF
--- a/src/components/app/DocumentList.js
+++ b/src/components/app/DocumentList.js
@@ -55,7 +55,7 @@ class DocumentList extends Component {
             viewId: defaultViewId,
             page: defaultPage || 1,
             sort: defaultSort,
-            filters: null,
+            filters: [],
 
             clickOutsideLock: false,
             refresh: null,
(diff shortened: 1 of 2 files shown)


teosarca:
Did a quick test but now nothing happens when i press apply (i.e. nothing sent to backend):
image

image

vko-online:
@teosarca Yes, sorry, my bad, now i've reverted the initial state value. Works as expected now

metasnw:
after testing the filters are applied but are not displayed correctly.
@teosarca shall we close the PR ?

vko-online:
There should be another Issue for UI bindings of Filter

teosarca:
@metasnw yup. basically the frontend was sending the right filters to backend but after that, in UI, the filters were cleared and just the last one used was kept.

PR #1163, opened 2017-08-29
Implements POC for intro/hints using intro.js #mf15-9

wiadev:
#mf15-9


wiadev:
please keep in mind following aspect:

1. Due to quite complicated flow to constantly append query parameter (in particular affects views, pagination etc) I'm using global application state once page loaded instead to replace URLs on every router navigation.

2. So to enable tutorial you should add following parameter into URL: ?tutorial=true or &tutorial=true (in case when URL already contains query parameters). Better to use query parameter with value (e.g. tutorial=true) instead of just ?tutorial. This is React router specifics

3. Further navigation will re-use this setting until you do FULL PAGE REFRESH (e.g. Ctrl-R or F5). If you want to continue with tutorial enabled - please append query parameter once again

4. If you want to forcefully disable tutorial - just perform page refresh

Examples:

Tutorial will be disabled:
http://localhost:3000/window/53100/2188396

Tutorial will be ENABLED for this page and further router navigation events:
http://localhost:3000/window/53100/2188396?tutorial=true

PR #1168, opened 2017-08-30
Allow inbox notification delete #1157

wiadev:
#1157


teosarca:
quick checked this pull request, but after i've deleted a notification the frontend called the "/all" API instead of listening on websocket "notification deleted" event.

image


The downside of not listening on websocket events is in case user has several browser tabs opened and in one of them he/she is deleting a notification event, the notification event won't vanish from all the other browser tabs.

teosarca:
quick tested, awesome! :)

PR #1187, opened 2017-09-07
Applies additional checks whether we need to scroll view #1183

wiadev:
Applies additional checks whether we need to scroll view to show element upon navigation in dropdown/lookup lists #1183


wiadev:
@teosarca ready

Issue #1060, opened 2017-07-19
Locale viewing and editing in User Interface


metas-mk:
### Type of issue
Feature request
### Current behavior
Currently, switching the Locale in Avatar does not affect the formatting in User Interface. 
For example in amount widget:
de_DE: 1.000,00 (tsd seperator '.' decimal sign ','
de_CH: 1'000.00(tsd seperator ''' decimal sign '.' 
en_US: 1,000.00(tsd seperator ',' decimal sign '.' 
and so on.
### Expected behavior
The locale set in Avatar shall respect the locale in User Interface too, so viewing and editing in locale is allowed and possible.

metas-mk:
Belongs to https://github.com/metasfresh/metasfresh/issues/2025

wiadev:
we format amount and costprice fields only
if you need more types - let me know
currently we have 4 locales hard-coded - de_DE, de_CH, nl_NL, nl_BE
formats taken from https://lh.2xlibre.net/locale/de_CH/

metas-lc:
IT
created a purchase order and added a big qty and price. Changed language and checked line net amt 
de_DE
3.273,60
de_CH
3'273.60
en_US
3,273.60

OK

PR #1193, opened 2017-09-11
Implements locale-specific formatting of numbers (amounts, prices) #1060

wiadev:
#1060

### DIFF
--- a/src/constants/Locales.js
+++ b/src/constants/Locales.js
@@ -1,0 +1,103 @@
+export const LOCAL_NUMERAL_FORMATS = {
+    'en_us': {
+        defaultFormat: '0,0.00',
+        delimiters: {
+            thousands: ',',
+            decimal: '.'
+        },
+        abbreviations: {
+            thousand: 'k',
+            million: 'm',
+            billion: 'b',
+            trillion: 't'
+        },
+        ordinal: function (number) {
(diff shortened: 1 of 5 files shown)


src/constants/Locales.js
```diff
@@ -0,0 +1,103 @@
+export const LOCAL_NUMERAL_FORMATS = {
```
teosarca: @wiadev  I have following questions regarding this file.
What's the reason for having those abbreviations there; i mean, where are they used?

IMHO, i think only the thousands separator and decimal separator are needed.

More, i think it would be better if those "locale" info would be provided in http://w101.metasfresh.com:8081/rest/api/userSession response.
wdyt?

teosarca: or we can even deliver the delimiters in layout element, i.e.

image

teosarca: update: agreed with Mark to create a followup issue

wiadev: This is requirements of external library we're using for formatting - numeral.js.

Please check it there:  
http://numeraljs.com/#locales

@teosarca

wiadev: I saw this. But it contains precision only. We don't use abbreviations, they're placed only according to requirements, but not used in fact.

Issue #1205, opened 2017-09-19
Can't open any process while on Dashboard


metas-lc:
### Type of issue
Bug 
### Current behavior
can't open any process while on Dashboard
### Expected behavior
process shall open
### Steps to reproduce
1. open https://w101.metasfresh.com:8443/
2. go to navigation
3. start a process (any process). e.g. Initial Setup Wizard
=> nothing happens

metas-ts:
cherry-picked to `release` branch

metas-lc:
IT
repeated the testcase above => works fine
also tried with opening a process from other window or from sitemap => the process starts OK

PR #1213, opened 2017-09-21
Fix for dashboard being stuck in infinite loop

PabloSichert:
#1205


PabloSichert:
Looks like this issue was introduced in commit 795c306. Calling `this.setState` leads to calling `componentDidUpdate` in the next cycle, causing it to be stuck in an infinite loop when there is no condition to abort.

@wiadev could you have a look at that commit? Did you mean to use `componentDidMount`?

PR #1282, opened 2017-10-19
Fix opening an includedView after process creation

ottosichert:
(See #1258)

@teosarca @metas-mk Please try out opening and closing all kind of modals and dialogs. Also you should notice that this PR fixes some minor issues such as losing selection and closing includedViews unintentionally when changing Table selection and back (works now)


ottosichert:
I tried following steps:

- Open **Picking Terminal** (https://w101.metasfresh.com:8443/window/540345)
- Select first item
- Click quick action `Pick`
- Modal opens
- First item in table and includedView is already selected
- Click quick action `Open HU selection window`
- Select a TU in new includedView
- `Uncaught Error: InvalidStateError: The connection has not been established yet`

_I this supposed to work?_

- Open **Picking Terminal** (https://w101.metasfresh.com:8443/window/540345)
- Select first item
- Click quick action `Pick`
- Modal opens
- First item in table and includedView is already selected
- Click drop down of quick actions
- Click `Pick to new HU`
- Select item in `Packvorschrift-Produkt Zuordnung`
- Click `Start`
- Selected item appears below item in includedView

_Is this correct?_

src/reducers/windowHandler.js
```diff
@@ -87,26 +86,17 @@ export default function windowHandler(state = initialState, action) {
...
-                    isAdvanced: false,
-                    title: '',
-                    type: '',
-                    modalType: '',
-                    saveStatus: {},
-                    validStatus: {},
-                    includedTabsInfo: {},
-                    triggerField: null
-                })
+                modal: Object.assign({}, state.modal, initialState.modal),
```
PabloSichert: Better: use

```js
modal: {
    ...state.modal,
    ...initialState.modal
}
```

We can do a refactoring specifically to replace all `Object.assign`s.

ottosichert: I kept it for consistency, but you are right we should use this in all occurences instead as we use babel's `object-rest-spread` already

PR #1299, opened 2017-10-25
Implement "DateRange" widget

PabloSichert:
#981 

Can someone explain the difference between those cases and where exactly they are used within the system?

Edit: for more context
```js
switch (widgetType)
```
/Edit

```js
case 'Date' && range
```

```js
case 'DateTime' && range
```

```js
case 'DateRange'
```

---

See the redundancy which I would like to resolve:

https://github.com/metasfresh/metasfresh-webui-frontend/blob/89198c89f87d7f0c7c3bd76bbb8d4086a7a3fa9c/src/components/widget/RawWidget.js#L202-L228

https://github.com/metasfresh/metasfresh-webui-frontend/blob/89198c89f87d7f0c7c3bd76bbb8d4086a7a3fa9c/src/components/widget/RawWidget.js#L257-L284

https://github.com/metasfresh/metasfresh-webui-frontend/blob/89198c89f87d7f0c7c3bd76bbb8d4086a7a3fa9c/src/components/widget/RawWidget.js#L312-L329

### DIFF
--- a/src/components/widget/RawWidget.js
+++ b/src/components/widget/RawWidget.js
@@ -309,6 +309,25 @@ class RawWidget extends Component {
                         </div>
                     )
                 }
+            case 'DateRange': {
+                return (
+                    <DatetimeRange
+                        onChange={(value, valueTo) => (
+                            this.handlePatch(widgetField, {
+                                ...(value && { value }),
+                                ...(valueTo && { valueTo })
+                            })
+                        )}
+                        mandatory={widgetData[0].mandatory}
+                        validStatus={widgetData[0].validStatus}
+                        onShow={onShow}
+                        onHide={onHide}
+                        value={widgetData[0].value}
+                        valueTo={widgetData[0].valueTo}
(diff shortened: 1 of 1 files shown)


metas-mk:
@PabloSichert 
Here are some examples for your questions above.

'DateTime' && range (based on fields/ columns which are defined as Dates with additional time information (similar to a Timestamp)).
image

'Date' && range (same as above but without Time).
image

I have no idea until now why there is an additional 'DateRange' case.

PabloSichert:
Thanks for clearing up. Seems like `case 'Date' %% range` and `case 'DateRange'` should be handled the same. I think `DateRange` was newly added to the API, but I'm not sure. Is this an unintentional duplication on the backend? @teosarca

teosarca:
@PabloSichert  for the case when it's Date and it's range i've introduced a new widget type called DateRange. I did that because also because the data type exchanged by this widget (i.e. {value:"", valueTo:""}) is different from a regular Date widget.

So, yes, they should be handled the same.

PR #1309, opened 2017-10-30
Processes: provide current selected tab and rows

PabloSichert:
#1295 

Note: This breaks #1279, but has cleaner code to continue with.


PabloSichert:
Oh - hah, yes 😄

PR #1392, opened 2017-11-29
new getAttributes methods #1357

wiadev:
#1357

### DIFF
--- a/src/actions/GenericActions.js
+++ b/src/actions/GenericActions.js
@@ -39,21 +39,20 @@ export function getData(
 }
 
 /**
- * 
- * @param {*} entity - entity should always be `documentView`
- * @param {*} windowId 
- * @param {*} viewId 
- * @param {*} rowId 
+ *
+ * @param {*} windowId
+ * @param {*} viewId
+ * @param {*} rowId
  */
-export function getAttributes(
-    entity, windowId, viewId, rowId
+export function getViewAttributes (
+    windowId, viewId, rowId
 ) {
     return axios.get(
         config.API_URL +
-        '/' + entity +
+        '/documentView'+
         '/' + windowId +
-        (viewId ? '/' + viewId : '') +
-        (rowId ? '/' + rowId : '') +
+        '/' + viewId +
+        '/' + rowId +
         '/attributes'
     );
 }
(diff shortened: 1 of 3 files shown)


src/actions/GenericActions.js
```diff
@@ -38,6 +38,26 @@ export function getData(
     );
 }
 
+/**
+ * 
+ * @param {*} entity - entity should always be `documentView`
```
teosarca: then why having it as parameter?

src/actions/GenericActions.js
```diff
@@ -38,6 +38,26 @@ export function getData(
...
+ */
+export function getAttributes(
+    entity, windowId, viewId, rowId
+) {
+    return axios.get(
+        config.API_URL +
+        '/' + entity +
+        '/' + windowId +
+        (viewId ? '/' + viewId : '') +
+        (rowId ? '/' + rowId : '') +
```
teosarca: imho the viewId and rowId shall always be present. if not, there is an error.

src/actions/GenericActions.js
wiadev: just for the future development

src/actions/GenericActions.js
wiadev: yes, they will always be presented but left them to make it generic function

wiadev: @teosarca do you want me to update these comments?

teosarca:
found an issue: wrong endpoint is called when PATCHing:
image

src/actions/GenericActions.js
teosarca: Agile development best practices(1) says that we shall not develop ahead and imagine things because they might never happen. It might be that there will be no future extension. And when and if it will be, we will address it at that time.
At the moment i find this misleading for new developers.
More I would rename the function from "getAttributes" to "getViewAttributes" because that's what precisely is doing.
I will drop that "entity" param because it;s pointless and i will also drop that viewId and rowId checking because if they are null/missing, you would build a wrong endpoint URL anyways.

Also, please note, I am not a javascript developer, so it might be wrong what I am saying.
I would like to hear other oppinions too.
@PabloSichert  @metas-ts  @ottosichert  guys, wdyt?

References:
1. Clean Code: A Handbook of Agile Software Craftsmanship (Robert C. Martin)

Issue #1348, opened 2017-11-16
Keeping lines selected when turning the page not working


metas-dh:
### Is this a bug or feature request?
Bug
### What is the current behavior?
When you select lines in grid view, and turn the pages, the lines don't stay selected.
#### Which are the steps to reproduce?
1. Open e.g. Handling Unit Editor
2. On 1st page, select all lines in all pages with 2x [ctrl]+a
3. Turn to page2: nothing selected, NOK
4. Turn back to page 1: nothing selected, NOK
### What is the expected or desired behavior?
Lines shall remain selected when you turn the pages.

metas-dh:
Results of IT1
tested in mf15

* selecting all lines in all pages in grid view, turning the page (using left & right key, page up & page down shortcut, or selecting the page directly): all lines are still selected in all pages, OK
* selecting all lines in 1 page in grid view, turning the page (using left & right key, page up & page down shortcut, or selecting the page directly): all lines are still selected in this page, OK

PR #1458, opened 2017-12-15
Remove deselect feature from TablePagination #1348

wiadev:
- Add prototypes

#1348


metas-mk:
Quick Test done in dev-branch. OK.

PR #1464, opened 2017-12-16
Prevent propagation of keydown event from Datetime #1429

wiadev:
- Add propTypes

#1429


metas-mk:
Quick Test before merge done. OK.

PR #1497, opened 2018-01-10
Fix drawing bars for negative values #1427

wiadev:
#1427


metas-mk:
Quickly checked locally in dev-1427 branch. Works now. Integrating.
image

Issue #1495, opened 2018-01-10
Error on "add to transportation order"


metas-lc:
### Is this a bug or feature request?
bug
### What is the current behavior?
error when you use "add to transportation order"
#### Which are the steps to reproduce?
1. open Picking Tray Clearing (Prototype) window/540371
2. select an LU from a slot (left side) and use Take out HU
=> now your HU is on the right side
3. select "add to transportation order" action from right side
Error: Could not read JSON...
Client error
Could not read JSON document: Can not deserialize instance of java.util.HashSet out of VALUE_STRING token at [Source: java.io.PushbackInputStream@2eb43e70; line: 1, column: 209] (through reference chain: de.metas.ui.web.process.json.JSONCreateProcessInstanceRequest["parentViewSelectedIds"]); nested exception is com.fasterxml.jackson.databind.JsonMappingException: Can not deserialize instance of java.util.HashSet out of VALUE_STRING token at [Source: java.io.PushbackInputStream@2eb43e70; line: 1, column: 209] (through reference chain: de.metas.ui.web.process.json.JSONCreateProcessInstanceRequest["parentViewSelectedIds"])

(workaround: select another picking slot, select your slot back, select your HU, use add to trasnportation)
### What is the expected or desired behavior?
shall open the process, no errors

metas-lc:
IT
checked both testcase and workaround => works fine: the error doesn't appear anymore.  "add to transportation order" action opens OK

PR #1511, opened 2018-01-18
Make sure parentViewSelectedIds is an array in process payload #1495

wiadev:
#1495


metas-mk:
Quick check in a local environment. Integrating.

Issue #1509, opened 2018-01-18
Harmonize details calendar widget with daterange widget in Filter


metas-mk:
### Is this a bug or feature request?
Feature Request
### What is the current behavior?
Currently, the overlay of the date widget in filter dies not harmonize exactly with the look and feel of daterange widget.
#### Which are the steps to reproduce?
Open and see.
### What is the expected or desired behavior?
Pls do the following improvements:
- Add a triangle pointer to date widget when opened
- reduce the gap between date field and widget

![image](https://user-images.githubusercontent.com/13365687/35083100-9e2c056e-fc1d-11e7-8610-ed399e7b649e.png)

![image](https://user-images.githubusercontent.com/13365687/35083108-a954b864-fc1d-11e7-8899-e04aa65f5556.png)

metas-lc:
IT
checked both filters: date and date promised in order window
=> both have triangle pointer and same small distance between date field and widget OK

PR #1513, opened 2018-01-18
Add a triangle pointer to date widget #1509

wiadev:
#1509


metas-mk:
Quick check in a local environment. Works nicely. Integrating.

Issue #1252, opened 2017-10-12
Select all x items shall not be available when there is only one page of entries


metas-rc:
### Is this a bug or feature request?
* FR
### What is the current behavior?
* When having only 1 page of entries in a window we have the option to select all x elements after selecting all on the page.
#### Which are the steps to reproduce?
* Go to a window with only 1 page on entries ( I had only one page in Handling Units Editor, so this was my example)
* There is the option to select all entries from a page: OK
* After selecting all, there is the option to select all x items ( in my case there were 13 HUs) -> NOK: This is irrelevant and shall not be when we only have 1 page of entries.
### What is the expected or desired behavior?
* The Select all x items shall not be available when we have only 1 page of elements.

metas-rc:
Followup for https://github.com/metasfresh/metasfresh-webui-frontend/issues/1221

metas-lc:
IT
checked Handling units window, first only with 1 page of entries
=> select all x items option didn't appear after you selected all OK
then with 2 pages of entries
=> select all x items option appear after you selected all OK

select all x items option worked when chose (select all on this page option appeared) OK
checked in other documents - worked OK

PR #1524, opened 2018-01-22
Hide Select all x items if there is a page only.

wiadev:
#1252


metas-mk:
Quick check. Works nicely.

Issue #1528, opened 2018-01-24
Added subrow is not shown in the frontend


metas-ts:
### Is this a bug or feature request?
Bug
### What is the current behavior?

Note that this issue can be ovserved in the described way as soon as the current WI for 
[metasfresh#3353](https://github.com/metasfresh/metasfresh/issues/3353) Implement MSV3 availability query and purchase order 
is integrated

In the SO2PO model overlay view, some sub-rows may be added after the initial loading, in an asynchronous fashion.
However, they are not shown.

On the following screenshot, i highlighted the place where i expect the row (left) and the JSON data which contains the JSON data I expect to be displayed in that row.
Note that the JSON data was queried by the frontend due to a websocket notification form the API
![image](https://user-images.githubusercontent.com/1264789/35327642-6d624984-00fa-11e8-9e77-44878308ef27.png)

PabloSichert:
Hey @metas-ts, could you provide some more detailed instructions on how to get to this state?

I get a 404 for `/window/143/1000005`. Tried with other existing sales orders, but could not reproduce.

metas-mk:
@PabloSichert 1000005 is a testcase that is only on tobi's local database. just drop that and select any sales order w/ "convenience salad e.g. https://w101.metasfresh.com:8443/window/143/1000360.

PabloSichert:
Thanks, got it now

metas-ts:
Hi @PabloSichert ,
as @metas-mk wrote:
* on w101, e.g. go to https://w101.metasfresh.com:8443/window/143/1000360., or create a new sales order with at least one order line of "P002737 Convenience Salat 250g" (no need to complete it though)
* select the orderline and run the "Create purchase orders" action, like here:
![image](https://user-images.githubusercontent.com/1264789/35495703-806fb07c-04c2-11e8-86c6-bff93bb316ae.png)

metas-mk:
@PabloSichert a current example in w101
![image](https://user-images.githubusercontent.com/13365687/35510482-8a2348c8-04f8-11e8-9ca0-f65bcd5d9694.png)

metas-lc:
IT
1. create a MSV3 Lieferanten Konfig, for bp X
2. create a new sales order with a product that has multiple bpartner products, one of them bp X
3. select the line, go to actions
4. open Create purchase orders
=> notice that on the line with bp X you have another included line OK
[...]

PR #1551, opened 2018-02-02
Display asynchronously loaded sub-rows

PabloSichert:
#1528


metas-mk:
Quick test locally. Integrating.

Issue #1552, opened 2018-02-02
Selecting country in dropdown with mouse click not possible


metas-dh:
### Is this a bug or feature request?
Bug
### What is the current behavior?
Selecting a country with mouse click in the dropdown is not possible.
#### Which are the steps to reproduce?
1. Try setting the country in a bpartner location: start writing e.g. "deut"
=> Deutschland is displayed in the Dropdown
2. Try selecting it with mouse click:
=> not possible, you're asked if you want to leave the page: NOK
(working fine using Enter though)
### What is the expected or desired behavior?
Should work.

metas-lc:
IT
repeated the testcase above: works fine. You can add the country with mouse click (also tried using the keyboard)
it isn't asking me if i want to leave the page, only if i leave the country empty OK

short regression: checked other search fields in docview and subtabs => you can select them with mouse OK

PR #1556, opened 2018-02-03
Fix List and Lookup dropdown mouse event handlers

siemiatj:
This is a fix for #1536 and #1552 together in separate commits since they're very similar and happening in the same components.

Widget for changing attributes now allows multiple changes, and saves them after clicking outside of the widget. #1536

Lookup dropdown allows selecting option with mouse. #1552


siemiatj:
Codeclimate is complaining but I've checked the issues and nothing major there.

PR #1560, opened 2018-02-06
Fix errors in the HU Editor window

siemiatj:
Not in all cases we're sending `fields` data with the request, so the UI should not trip over if this happens. 

Plus a small cleanup, rewriting this function to leverage ES6 spread operator.


metas-mk:
Belongs to https://github.com/metasfresh/metasfresh-webui-frontend/issues/1559

metas-mk:
Quickly tested locally. Describes errors don't pop up anymore. Integrating.

Issue #1553, opened 2018-02-02
Drop down lists remain displayed when using tab several times (again)


metas-dh:
### Is this a bug or feature request?
Bug
### What is the current behavior?
pls see #1421, where this was fixed before.

metas-lc:
IT
1. created a sales order using tab and keyboard 
=> works fine OK
2. used tab multiple times
=> Drop down lists close each time OK

PR #1563, opened 2018-02-06
Fix mouse and keyboard events handling in RawList component

siemiatj:
There was a huge mess in how we dealt with events. So I cleaned this up and removed half of the variables responsible for setting the state of the RawList component. Seems to be much more stable and easier to maintain now.

### DIFF
--- a/src/components/app/QuickActions.js
+++ b/src/components/app/QuickActions.js
@@ -47,10 +47,6 @@ class QuickActions extends Component {
     }
   }
 
-  componentDidMount = () => {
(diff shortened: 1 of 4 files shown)


teosarca:
belongs to https://github.com/metasfresh/metasfresh-webui-frontend/issues/1553

metas-mk:
Testing Locally. Dropdowns vanish, but I get this
image
<-- changed fields yellow does not vanish
<-- cursor is focused in url field instead of metasfresh invoicepartner/ warehouse

Issue #1501, opened 2018-01-14
Automated Javascript and React Testing with Jest


metas-mk:
### Is this a bug or feature request?
Evaluation
### What is the current behavior?
The Testing Framework used by Facebook to Test Javascript and React applications is Jest. Check out further Information her https://facebook.github.io/jest/
It seems easy to setup and has a cool feature called snapshots, which allow to ease the creation of valid results for the tests, in environments with fast changing and variable data. This could be a nice feature that we need for our automated testing based on data that can change in database in a short timeframe.
Additionally its said that Jest is 3 times faster than Mocha in Test execution.
#### Which are the steps to reproduce?
None.
### What is the expected or desired behavior?
Check and evaluate jest for the testing of metasfresh-webui-frontend.

Install Jest in metasfresh-webui-frontend:
```yarn add -D jest```

Run Jest in watch mode:
```yarn jest --watch```
then start test run for all current tests with ```a```

metas-mk:
https://github.com/metasfresh/metasfresh-webui-frontend/pull/1569

metas-rc:
Nothing to test. Moving to done.

PR #1569, opened 2018-02-08
Improve the test environment

siemiatj:
Decided to use facebook's Jest for our testing environment, as it's really easy to work with, and gives a lot of built-in options (like test coverage we can turn on later) and flexibility. Rewritten current tests as well  so everything should work in Jenkins too.

https://github.com/metasfresh/metasfresh-webui-frontend/issues/1501


metas-ts:
hi @siemiatj i would like to help making them execute on jenkins.
currently, we basically do this:
```bash
yarn install
yarn lint --quiet
yarn test
webpack --config webpack.prod.js --bail --display-error-details
```
in the Jenkinsfile...

siemiatj:
@metas-ts It'll run on Jenkins in the same way as it did before. We just need more tests that actually test something. What we had right now was barely testing anything so they'd never crash :)

Issue #1566, opened 2018-02-06
Subtab Data not shown when clicked first time


metas-mk:
### Is this a bug or feature request?
Bug
### What is the current behavior?
Currently, when opening a Subtab, the data is not shown when the first time called/ clicked.
#### Which are the steps to reproduce?
- Open window/123 for Business partner "Machine Parts Ltd"
- Click on Subtab Customer: empty
- Click on Subtab Vendor: empty
- Click back on Subtab Customer: filled now
### What is the expected or desired behavior?
The Subtab Data shall be filled already when called/ clicked the first time and has data.

![image](https://user-images.githubusercontent.com/13365687/35868610-4d0f82d8-0b5d-11e8-8fab-037f0e7207d4.png)

metas-mk:
Was broken with this PR https://github.com/metasfresh/metasfresh-webui-frontend/pull/1551

siemiatj:
Ok, this will be my first thing for tomorrow.

metas-ts:
cherry-picked into `release`

metas-lc:
IT
checked subtabs in business partner, order, shipment schedules
=> data is shown OK

PR #1573, opened 2018-02-08
Subtabs not showing data on first visit

siemiatj:
The comparison between the old vs new data was not checking deep enough in the data structure.

This is a quick fix, but the whole Table component needs a serious rewrite. We're doing tons of unnecessary things, requests etc there. I will have to come up with a plan how to leverage redux store for this too.

Relates to #1566

### DIFF
--- a/src/components/table/Table.js
+++ b/src/components/table/Table.js
@@ -128,7 +128,7 @@ class Table extends Component {
       );
     }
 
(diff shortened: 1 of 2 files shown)


metas-mk:
Tested locally. Works nicely. Integrating.

Issue #1554, opened 2018-02-02
Action Parm modal Overlay focus missing n 1st Field


metas-mk:
### Is this a bug or feature request?
Feature Request
### What is the current behavior?
Currently, the Modal Overlays for Process/ Action Parms don't have the focus on the first recordable field. See example in Picking Terminal described below.
#### Which are the steps to reproduce?
- Open https://w101.metasfresh.com:8443/window/540345
- select 1st line Ananas and start "Pick" Action
- Action "Pick to new HU"
- Press ```tab``` multiple times. No focus recognizable.
### What is the expected or desired behavior?
The focus shall always be on 1st recordable fields in Process/ Action Pam modal overlays.

![image](https://user-images.githubusercontent.com/13365687/35738784-28f4b9f2-0830-11e8-9e2c-b769b13fab63.png)

siemiatj:
Tested my PR changes as described in the issue + on `/window/143/1000386`.

metas-dh:
Results of IT1
tested in mf15

focus shall always be on 1st recordable fields in Process/ Action Pam modal overlays:

* repeated testcase above: focus is in the 1st field: OK
* checked other Process/ Action Pam modal overlays as well (in picking, manufacturing, material receipt etc.): focus is in the 1st recordable field: OK

Issue #1550, opened 2018-02-02
Dropdown List Widget does not show 1st line as selected with keyboard sroll


metas-mk:
### Is this a bug or feature request?
Bug
### What is the current behavior?
When in Subtab Data structure, and double-clicking on a Dropdown List widget, then scrolling down with ```arrow-down``` then the first entry is not selected. Same when moving up again.
#### Which are the steps to reproduce?
- Open Business Partner window window/123/2156438
- Go to Vendor Subtab
- Double Click on Field Purchasing Price System
- Press ```arrow-down```
### What is the expected or desired behavior?
Shall show the first entry as selected too.

![image](https://user-images.githubusercontent.com/13365687/35719053-a1f6bb8c-07e7-11e8-894c-acd4c44f4ff5.png)

siemiatj:
Do we want to indicate somehow that this thing is in the release already ?

PR #1576, opened 2018-02-09
Fix auto focus for List inputs in modal and selecting first option on the list

siemiatj:
Param name was wrong. I'm not sure how this was working before but this solves #1554 . Also refactored List and RawList components, decoupling them, extracting some code to a separate function (which will make this easier to test) and moving focus/toggle control to parent component. Now it should be easier to debug and understand what's going on when events are not flying around like crazy.
This also changes how select fields behave when selected with keyboard. No by default they are focused (which a visual indication) first and require user to click arrow down to show the list. Lookup fields work as before.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -16,7 +16,7 @@
           "singleQuote": true,
           "trailingComma": "es5",
           "bracketSpacing": true,
-          "jsxBracketSameLine": true,
+          "jsxBracketSameLine": false,
           "parser": "flow"
         }],
         "react/prop-types": "warn",
--- a/package.json
+++ b/package.json
@@ -80,7 +80,8 @@
     "stylelint": "^7.9.0",
     "webpack": "~1.13.0",
     "webpack-dev-server": "~1.12.1",
-    "webpack-node-externals": "^1.6.0"
+    "webpack-node-externals": "^1.6.0",
+    "why-did-you-update": "^0.1.1"
   },
   "dependencies": {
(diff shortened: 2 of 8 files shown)


siemiatj:
Don't look at codeclimate. Nothing interesting there.

metas-mk:
Checked locally. Using keyboard only in the first step.

- In Picking modal overlay, Choose ```alt+l``` to open the quickaction list. select ```Pick to new HU```.
image

- Next thing the user sees is a focussed List widget (OK) but the dropdown list is expanded. I believe it shall be focused but collapsed for such kind of action parms. The user can press ```space``` to expand. Also when expanding the dropdown the first entry shall be visible selected initially.
image

- After selecting an entry, confirming with ```enter```, the same dropdown list was opened another time. Shall not be.

@siemiatj pls ping me if we shall discuss the options.

siemiatj:
I think selecting the first element in dropdown is covered by : https://github.com/metasfresh/metasfresh-webui-frontend/issues/1550

metas-mk:
Checked locally. Boxes around lookup widgets are missing. NOK
image

Dropdown lists are now not opening automatically when focused. Arrow down has to be pressed. OK

Dropdown in first Parm of action menu is still opening automatically. NOK
image

metas-mk:
Checked locally. Works as required. Minor issues found but will be addressed in follow-up issues.

Issue #1546, opened 2018-02-01
Date Range Filter as Range Filter has initial daterange


metas-mk:
### Is this a bug or feature request?
Bug
### What is the current behavior?
Currently, the Date Range Filter as Range Filter has an initial Range set when opening the filter. This makes it impossible to search without the Date Range Filter set properly.
#### Which are the steps to reproduce?
- Open Manufacturing Order window /window/53009
- Open "Filter"
- Check Preparation Date
### What is the expected or desired behavior?
Shall have an empty field initially. When selecting the Filter then set an initial Date in Daterange picker.

![image](https://user-images.githubusercontent.com/13365687/35667619-a84de7d0-072e-11e8-99df-5228adbce5d6.png)

metas-lc:
IT
1. manufacturing order - create some, with preparation date, some without
2. use preparation date filter
=> when you open the filters, preparation date is not set (See all dates) OK
3. set a date range
=> only those with preparation date in that range are shown OK
4. clear the filter
=> all documents are shown OK

PR #1577, opened 2018-02-09
No initial date in DateRange input

siemiatj:
Input for selecting DateRange now won't have initial date (when start/end date not provided).
Plus refactoring.

Relates to #1546 .


metas-mk:
Checked locally. Works nicely. Minor change needed. The String "All dates available" is not yet translated into other languages (e.g. de_DE)
image

Issue #1590, opened 2018-02-13
Included tabs sorting


metas-lc:
### Is this a bug or feature request?
bug
### What is the current behavior?
you can sort included tab only once and the arrow is not updating
#### Which are the steps to reproduce?
1. open sales order, create a new one and add 2 lines
(you can use this one https://w101.metasfresh.com:8443/window/143/1000425)
2. select Line No. to sort
=> rows are sorted OK
=> arrow is still up NOK
3. sort by Line No. again
=> nothing happens NOK 
### What is the expected or desired behavior?
each time you sort shall work
the arrow should update as well

metas-dh:
Results of IT1
tested in mf15

sorting in included tabs:
* rows are sorted: OK
* arrow is updated correctly: OK
* sorting works correctly as often as tried: OK

metas-ts:
cherry-picked to `release`

PR #1594, opened 2018-02-14
Fix sorting in table

siemiatj:
Fixed rendering and sorting in tables. Before we were really inefficient with that, plus we never stored the updated sorting order.

PR for #1590 

*EDIT (after review changes)*:

fix2

### DIFF
--- a/src/components/table/TableHeader.js
+++ b/src/components/table/TableHeader.js
@@ -1,47 +1,68 @@
-import cx from "classnames";
-import React, { Component } from "react";
+import React, { Component } from 'react';
+import classnames from 'classnames';
 
 class TableHeader extends Component {
   constructor(props) {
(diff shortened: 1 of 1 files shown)


metas-mk:
Checked locally. Sorting happens, but minor issues w/ sorting indicator
- after pressing a header column, the column shall be sorted by only that column and the indicator shall show if it's asc or desc sorting.
- after using manual sorting then only 1 column can have a sorting indicator. hide the indicator in columns that are not sorted.
- initially,/ per default sorted grid views can have more than 1 sorting columns and indicators. these are hidden as soon as user sorts manually (see above)

image

metas-mk:
Quick check locally. Works nicely as required. Integrating/ merging.

Issue #1593, opened 2018-02-14
Copy/ Paste for selected Field ub Subtab GridView not giving correct result


metas-mk:
### Is this a bug or feature request?
Feature Request
### What is the current behavior?
Currently, when selecting a row and field in Subtab Grid View and pressing [ctrl]+c, then the whole line content is copied.
#### Which are the steps to reproduce?
Open /window/540321/, try and see.
### What is the expected or desired behavior?
If the user explicitly selects a field, then only copy the content of the Field.
![image](https://user-images.githubusercontent.com/13365687/36198914-3d401540-1178-11e8-9f60-5db54bfab5bf.png)

metasnw:
one important remark: if the fields content is cut off the whole content is copied anyway
e.g. if an email address test@supercompany.com is displayed as test@sup... it still needs to be copied as test@supercompany.com  into clipboard

metas-dh:
Results of IT1
tested in mf15

* when selecting a row and field in Subtab Grid View and pressing [ctrl]+c only the field content is copied: OK
* if the fields content is cut off the whole content is copied anyway: OK

* when i had selected and copied a field, and then selected another row (in the same subtab or in the subtab in another window), and used [ctrl]+cthere, i again got the content of the field i copied first when pressing [ctrl]+v, not the content of the row i last selected & copied: NOK

question:
* when i had selected only the row and not a field, then i got not only the content of the row copied, but also the headers (e.g. "NameValid fromActiveDescriptionDiscount" etc. in PLV subtab): imho that's NOK, should be only the content of the selected row, right?

siemiatj:
@metas-dh Regarding the question : I'm not sure. It was working like that before.

siemiatj:
I tried reproducing this case you mentioned but it's working for me. Any tips how can I `break` it ?
![repro](https://user-images.githubusercontent.com/513460/36253981-a15fe042-1249-11e8-8e92-5e9712726b10.gif)

metas-dh:
pls try e.g. 
* copying the 0055 in the 3rd row
* then select the 1st row (select e.g. the 1st field of that row and then click somewhere else outside the rows, so that the whole row is selected, but not the 1st field)
* then do copy again: do u get the content of the whole 1st row or the 0055 again? (the latter happened for me)

metas-mk:
> then do copy again: do u get the content of the whole 1st row or the 0055 again? (the latter happened for me)
[...]

PR #1595, opened 2018-02-14
Copy only content of the cell with  Ctrl+c 

siemiatj:
Instead of massaging and merging row + headers data, we only want to copy content of the selected cell.

PR for #1593 

In action :
fix1

### DIFF
--- a/src/components/table/Table.js
+++ b/src/components/table/Table.js
@@ -749,49 +749,12 @@ class Table extends Component {
     );
   };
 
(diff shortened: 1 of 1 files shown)


metas-mk:
Checked locally. Works nicely as required. Will create follow-up issues for main grid view row selections. Integrating/ merging.

metas-mk:
Quick check with fields in edit mode. Is not copied.

PR #1597, opened 2018-02-14
Fix copying edited table cells

siemiatj:
The previous fix didn't handle the case when cell is edited. I'm not sure if it ever worked, but now seems to be returning the correct data. Depending on the field type we either take `textContent` or input's `value`.

Improvement for #1593


metas-mk:
Checked locally. Works nicely as required. Merging.

Issue #1592, opened 2018-02-14
Added subrow is still/again not shown in the frontend


metas-ts:
### Is this a bug or feature request?
Bug
From what i can see, issue
https://github.com/metasfresh/metasfresh-webui-frontend/issues/1528 Added subrow is not shown in the frontend
is back.
[This screenshot](https://user-images.githubusercontent.com/1264789/35327642-6d624984-00fa-11e8-9e77-44878308ef27.png) from the issue

### What is the current behavior?

If the asynchronous respons is coming back real quickly, then it is shown. Otherwise, it's still exactly the way it was described in issue #1528 .

#### Which are the steps to reproduce?

on w101:
* open a sales order with product P002737 ("Convenience Salad"), select the order line and open the "create purchase orders process.

on another machine, first mal sure that the "Test Lieferant" vendor has a MSV3 connection configured, that looks like this one: 
![image](https://user-images.githubusercontent.com/1264789/36195994-93f52258-116f-11e8-9c24-fe677ca6c16a.png)

(the URL is http://demsvtest:8089/msv3/v2.0 )

metas-ts:
if the issue is complicated, please ping me. the API could also invalidate "parent" row, instead of just the child row that actually got a new subrow

metas-lc:
IT
made config, created order, selected line and run create purchase orders process
=> line appears, readonly OK

PR #1599, opened 2018-02-15
Added subrow is still/again not shown in the frontend

siemiatj:
Websocket response mutates the rows data, and because they're passed by reference our test failed to notice any changes. I will see if we can use Immutable here to be safe from that in the future.

Some refactoring too.

PR for #1592


metas-ts:
changing reviewer from me to @metas-mk

metas-mk:
Quick check locally. MSV3 server did not respond. Row w/ IO error was shown. Integrating for Test on w101.

Issue #1598, opened 2018-02-14
Changes in location not always saved


metas-dh:
### Is this a bug or feature request?
Bug
### What is the current behavior?
When you add a location, and you add the country, then click e.g. street and then outside of the address modal window, the address is not saved; and when you close the modal you get a new, red line in location subtab (disappears after refresh).  You have to click outside of the modal window directly after setting the country or add e.g. street to have your changes saved.
#### Which are the steps to reproduce?
1. Set a country, click in street and then next to the address modal: 

![2018-02-14 18_02_52-metasfresh](https://user-images.githubusercontent.com/18696715/36217380-8fc717f8-11b1-11e8-9ff7-0c6913aaa890.png)

2. Close the modal with Done, OK:

![2018-02-14 17_47_24-metasfresh](https://user-images.githubusercontent.com/18696715/36217191-19b5963e-11b1-11e8-9c98-318273bcf16a.png)

3. Set country, click directly outside the modal:

![2018-02-14 17_51_27-metasfresh](https://user-images.githubusercontent.com/18696715/36217451-c143c8f8-11b1-11e8-9fa6-693f98336cd9.png)

### What is the expected or desired behavior?
Changes shall be saved.

metas-ts:
:cherries: to `release`

metas-dh:
Results of IT1
tested in weekly e2e check

location is always saved, no matter where you click or in which order, after setting the country: OK

PR #1602, opened 2018-02-15
Fix saving location changes

siemiatj:
Our current logic for handling showing/hiding/submitting of widgets is terrible. Here's a major rewrite relying less on savingstates in variables and switching between focused/blurred states.

For now I tested it in window 123 & 143 and looks to be working fine. ~Still need to add some tests.~

I'll add tests in a separate PR as it requires quite a bit of plumbing before.

### DIFF
--- a/src/components/app/Modal.js
+++ b/src/components/app/Modal.js
@@ -254,18 +254,18 @@ class Modal extends Component {
     dispatch(closeModal());
 
     if (!rawModalVisible) {
-      document.body.style.overflow = "auto";
+      document.body.style.overflow = 'auto';
     }
   };
 
   handleClose = () => {
     const { modalSaveStatus, modalType } = this.props;
 
-    if (modalType === "process") {
+    if (modalType === 'process') {
       return this.closeModal();
     }
 
-    if (modalSaveStatus || window.confirm("Do you really want to leave?")) {
+    if (modalSaveStatus || window.confirm('Do you really want to leave?')) {
       this.closeModal();
     }
   };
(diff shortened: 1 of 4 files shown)


siemiatj:
Jesus both codeclimate and codebeat are just useless...

metas-ts:
this PR belongs to #1598

metas-ts:
hi @siemiatj 

> Jesus both codeclimate and codebeat are just useless...

when writing java, codeclimate usually fails alltogether (probably the repo is too big)..codebeat "often" shows a result, but would need a lot of tuning to get rid of many false positives..
So they are not super useful to me either, but i often take at least a glimpse at them..

Having that said, if they annoy you in metasfresh-webui-frontend, we can kick them out.

metas-mk:
Checked locally. Found minor glitches but fixes the described. Merging.

siemiatj:
@metas-mk can you roughly describe the glitches ? I'm trying to write tests for those components so maybe we could write a test case covering this.

metas-mk:
@siemiatj yes, just minor one(s)
- Open the window/123/ and add a new location
- press on ```edit``` button for location
- don't record anything, instead press on space outside or inside modal
- a message appears ```Do you really want to leave?```. press OK.
- the locator edit closes.

I think that the user should not get the question ```Do you really want to leave?``` just when pressing outside/ inside modal with locator editor opened. The locator editor should just close without content.

siemiatj:
Oh, cool. Yeah, it was weird to me too, but I thought it's intentional. This'll be a quick fix.

Issue #1394, opened 2017-11-29
Quickactions not loaded after using barcode/ std. filter in HU editor in Manufacturing Order


metas-mk:
### Is this a bug or feature request?
Bug
### What is the current behavior?
Currently, the Quickactions are not loaded after using barcode/ std. filter in HU editor in Manufacturing Order.
#### Which are the steps to reproduce?
1. Open the window https://w101.metasfresh.com:8443/window/53009
2. Select the first manufacturing order with No. 630015
3. Press Quickaction Issue/ Receipt
4. Select the Row w/ Type "CO"
5. Press Quickaction "HU-Auswahlfenster öffnen"
6. Press the filter Button in the opened included view. Press Barcode. Enter 1000086.
--> the included view is filtered and the Quickactions dissappear. From now on the Quickactions are not loades anymore when switching different HU.

### What is the expected or desired behavior?
Shall work.

![image](https://user-images.githubusercontent.com/13365687/33395469-e6781a32-d545-11e7-830d-c3db3e78b169.png)

Here the result. The Quickactions are missing:
![image](https://user-images.githubusercontent.com/13365687/33395596-591e8134-d546-11e7-8338-d0fb1df44e7f.png)

metas-dh:
update:
e.g. Barcode scan of LU: actions are shown correctly afterwards, but not after Barcode scan of a TU or CU

metas-dh:
Results of IT1
tested in mf15

* Quickactions loaded after using barcode in HU editor in Manufacturing Order: worked fine after filtering for barcode of LU, TU & CU: OK
* Quickactions loaded after using  std. filter in HU editor in Manufacturing Order: worked fine after filtering for value of LU, TU & CU: OK

PR #1610, opened 2018-02-20
Fix and clean table selections

siemiatj:
Each nested table was given a new id, and because of that selections were not saved properly. This breaks quick actions and probably something else as well.

Relates to #1394


siemiatj:
Cleaning lint.

metas-mk:
Did a quick local test. Issue reported is fixed. Further detailed tests needed.

Issue #1585, opened 2018-02-12
Date recording with keys leads to error


metas-mk:
### Is this a bug or feature request?
Bug
### What is the current behavior?
When recording a date with the keyboard, instead of clicking it with the mouse in date widget, then an error occurs and a blank screen occurs. 
#### Which are the steps to reproduce?
- Open window/143
- Click on a date widget field and start typing the date.
### What is the expected or desired behavior?
Shall also work with the keyboard.

![image](https://user-images.githubusercontent.com/13365687/36097847-f212d08e-0ffc-11e8-99b8-b96c187908f7.png)

metas-dh:
Results of IT1
tested in mf15

* date recording with keys: no error, no blank screen: OK
* date recording with keys is not saved everytime: 
  * when entering e.g. date of provisioning in window/143 with keys the first time, the date is saved
  * but when changing the date with keys, the changes are not saved, no progress bar after Enter or clicking somewhere else; when you jump back to previous window and back to window/143, you see the date as it was first set, changes are lost: NOK

siemiatj:
@metas-dh any ideas if this second case is a new thing or something that slipped through testing before ?

metas-dh:
@siemiatj tbh i'm not sure; i think it's a new thing but i wouldn't bet on it.

metas-dh:
checked again w/ metas-mk: changes w/ keys are saved when pressing Tab, now also saved when clicking somewhere else: OK.

PR #1611, opened 2018-02-20
Don't call Moment functions on strings

siemiatj:
Simple fix. We can only call the `utc(true)` function which resets timezones on Moment objects.

Relates to #1585


metas-mk:
Quick local Test. Works nicely as required. Merging.

Issue #1604, opened 2018-02-16
Date Range Picker shows wrong day of the week


metasnw:
### Is this a bug or feature request?
bug

### What is the current behavior?
![image](https://user-images.githubusercontent.com/16079655/36297573-c6c6ff1e-12f3-11e8-9cfa-066c2f865d27.png)

#### Which are the steps to reproduce?
1. open any date ranger picker
1. compare week day 

https://w101.metasfresh.com:8443/window/232

metas-rc:
Results of IT

Tested on webui


* OK: The day of week in date anges is correct in each case I tried: 
** Fielters
** Fields in single view
** Fields in grid view
** Fields in included tabs

PR #1618, opened 2018-02-22
Fix days shown in the DateRange calendar

siemiatj:
We were using days of the weeks starting from Sunday on a calendar starting from Monday.

This will make #1604 go away.


metas-mk:
Checked locally. Works nicely as required.
image
Merging.

PR #1619, opened 2018-02-22
Show date using the same formatting in TableCells

siemiatj:
Right now we were showing different datetime formats in table depending on the field editable state. readonly cell was using our predefined format, so I changed the picker field to use the same.

Relates to #1567


metas-mk:
Checked locally. Works nicely as required.
image
Merging.

Issue #1620, opened 2018-02-22
List widgets with focus shall have dark grey color underline


metas-mk:
### Is this a bug or feature request?
Feature Request
### What is the current behavior?
Currently, it is not recognizable very good for the user if the list widget is focused. This is because the focus underline has the same colors as other box borders or text underlines.
#### Which are the steps to reproduce?
Open, try and see.
### What is the expected or desired behavior?
Make the underline of focused list widgets in a dark grey color.

Followup of: https://github.com/metasfresh/metasfresh-webui-frontend/issues/1554

Example:
![image](https://user-images.githubusercontent.com/13365687/36540278-0cb5e00e-17da-11e8-9c4a-0948ac416bd5.png)

metas-lc:
IT

1. repeated the testcase above (picking terminal - pick to new hu)
=> the focus is on the first field OK
=> the field has a dark grey under it (while is focused) OK
=> select the second field: focus on the first one is not lost OK?
=> after the focus is lost, there is no line under packing item (only under qty) OK? 
=> pressing multiple tabs won't get me back to packing item OK?


other window. 
opened manufacturing order, there are 2 list widgets there
set all data
=> the dropdown fields have a dark grey under then (while are focused) OK
=> if you use tab to go from one field to another the focus disappears when you press tab OK
=> if you use mouse click and jump over a focused dropdown, that field remains focused (you 'll have 2 focused dropdowns at the same time) OK?
![image](https://user-images.githubusercontent.com/16001666/36788139-764922f0-1c95-11e8-81a8-545e06ff365c.png)
=> pressing multiple tabs got me back to beginning OK
=> after the focus is lost, there is no line under the field OK?

PR #1626, opened 2018-02-26
Darker underline for focused list widgets

siemiatj:


metas-mk:
Checked locally. Color is now much better to see the focused field. Merging.

metas-mk:
Belongs to https://github.com/metasfresh/metasfresh-webui-frontend/issues/1620

Issue #1631, opened 2018-02-26
Attributes not displayed for first line in material receipt candidate HU Editor


metas-dh:
### Is this a bug or feature request?
Bug
### What is the current behavior?
When you open the HU Editor for material receipt candidates, the first line is selected, but the attributes on the right side are not shown. You have to select the 2nd line and then the 1st line again to have the attributes displayed, clicking on the 1st line or clicking somewhere else in the window and then selecting the 1st line again has no effect=> when you have only 1 line, i.e. when you receive only CUs, you cannot set attributes!

![2018-02-26 17_28_17-metasfresh](https://user-images.githubusercontent.com/18696715/36682302-08a0875a-1b1b-11e8-872e-a1ed7d2254cc.png)

![2018-02-26 17_28_37-metasfresh](https://user-images.githubusercontent.com/18696715/36682312-10382ac2-1b1b-11e8-95f6-f3f22d44ba32.png)

#### Which are the steps to reproduce?
Try and see.
### What is the expected or desired behavior?
Attributes on the right side shall be displayed right away.

metas-ts:
:cherries: to `release` (as #1625 )

Issue #1625, opened 2018-02-26
Show Attribute Editor for selected CU Level HU


metas-mk:
### Is this a bug or feature request?
Bug
### What is the current behavior?
Currently, the Handling Unit Editor does not show the Attribute Editor when opening the Handling Unit editor with a selected Customer Unit (without box, palette)
#### Which are the steps to reproduce?

- Open Material Receipt Window https://w101.metasfresh.com:8443/window/540196
- Select a line and action "Receive CU's". HU Editor opens and the Attribute Editor is empty.
![image](https://user-images.githubusercontent.com/13365687/36654864-7d07acd8-1abf-11e8-9eb6-1affebbd594b.png)

When doing the same with 2 selected lines in step 1, and you switch between both lines in HU Editor, then the Attribute Editor appears after first switch.
![image](https://user-images.githubusercontent.com/13365687/36654880-9f178ca8-1abf-11e8-9a92-95cecbba5b1e.png)

### What is the expected or desired behavior?
The Attribute Editor shall be shown initially in Handling Unit Editor when a row (left view) is selected.

metas-lc:
IT
1. open hu editor with 1 CU line
=> attributes are visible OK

Regression
2. open hu editor with LU-TU-CU 
=> attributes are visible OK
3. open hu editor with TU-CU 
=> attributes are visible OK

metas-ts:
:cherries: to `release`

Issue #1623, opened 2018-02-24
Attributes not shown in Material Receipt Candidates


metas-mk:
### Is this a bug or feature request?
Bug
### What is the current behavior?
Currently, when opening the Material Receipt Candidates HU Editor with action "Receive CUs", then the Attributes for a selected line are not shown.
#### Which are the steps to reproduce?
Open, try and see.
### What is the expected or desired behavior?
The Attributes shall be visible for selected Customer units.

![image](https://user-images.githubusercontent.com/13365687/36631197-10ca676e-1974-11e8-8a23-ae76f44d903e.png)

metas-lc:
IT
tested in https://github.com/metasfresh/metasfresh-webui-frontend/issues/1625 OK

PR #1632, opened 2018-02-26
Fix loading attributes editor

siemiatj:
Regression bug. Looks like fixing #1394 broke loading attributes edit in HU editor. This PR introduces a mixed solution, using values from the store to trigger React's lifecycle events, but still has a fallback to update viewId (so that quickactions are properly displayed).

Related to #1623 #1625 #1631 .
Doesn't reintroduce #1394.


metas-mk:
Quick check locally in Material Receipt Candidates + HU Editor and Manufacturing Order. Works nicely. Merging.

Issue #1630, opened 2018-02-26
Setting attributes in orderline not working


metas-dh:
### Is this a bug or feature request?
Bug
### What is the current behavior?
When you set an attribute in orderline, you can select it, but is not set and saved in the orderline.
#### Which are the steps to reproduce?
1. Create an orderline (purchase or sales side) w/ a product that has asi set
2. Open the window for attribute selection
3. Select e.g. AdR from the dropdown
4. Click somewhere else to have your change saved: nothing happens
5. When you refresh the page, the attribute is gone

![2018-02-26 17_24_11-metasfresh](https://user-images.githubusercontent.com/18696715/36681915-2e0d300c-1b1a-11e8-8977-58e7a4924530.png)

### What is the expected or desired behavior?
Attribute shall be set in orderline as selected.

siemiatj:
Fix ready, will push after doing more tests in other parts of the application.

metas-rc:
Results of IT

Tested on webui

* OK: The attributes can be set and are correctly saved in the db.

PR #1633, opened 2018-02-27
Fix setting values in attributes dropdown

siemiatj:
Regression introduced when fixing attributes in address modal. This is a fix for #1630 and checked to not bring #1598 back.

~Adding tests.~


metas-mk:
Checked locally. Setting Attributes in Sales Order works now (besides the other issue with list widget and only 1 entry, not able to select "none" then). OK
Checked location editor: Changes in the location are not saved. Example http://localhost:3000/window/123/2155894. Open the Location Editor, Change the Street, Click outside. Save indicator is shown but the location is not updated. NOK.

siemiatj:
Yes, looks like it submits when changing address but only when clicking inside the attributes modal.

siemiatj:
@metas-mk ready to test

Issue #1628, opened 2018-02-26
Clicking in field w/ 1 value in dropdown sets the value directly


metas-dh:
### Is this a bug or feature request?
Bug, i think
### What is the current behavior?
If you have a dropdown w/ only 1 value in a field, e.g. pricelist in bpartner window, then clicking or double clicking in that field just sets the value, the dropdown does not open and you cannot properly select the value. Likewise, you cannot set the value back, since you don't get the "empty" option from dropdown.
#### Which are the steps to reproduce?
1. Open bpartner window
2. Select a bpartner that is not customer yet, make sure you have only 1 customer pricelist
3. Set customer=Y in the resp. tab, and double click in pricelist field:
=> customer pricelist is set directly, no dropdown opens for you to select
4. Try setting the value back to no pricelist: 
=> not possible, dropdown does not open
### What is the expected or desired behavior?
Value shall be set after selecting it from dropdown, setting it back shall be possible.

siemiatj:
Related task #1622

metas-rc:
Results of IT

Tested on webui

I  repeated the test case from the issue and it works fine: 
* OK: The pricelist is set directly, but I can change it
* OK: I have the none option
* OK: I can change between the pricelist and the none option as many times I tried.
* OK: The entry is saved correctly when I set the pricelist and when I set no pricelist.

metas-ts:
:cherries:  to `release`

metas-rc:
* Checked briefly in other windows. It seems to work well there, too.

metas-rc:
Some other problems found while testing
* deleting selected value is not possible. Followup: https://github.com/metasfresh/metasfresh-webui-frontend/issues/1648

* Scrollbar doesnt'work with mouse: https://github.com/metasfresh/metasfresh-webui-frontend/issues/1605

Issue #1622, opened 2018-02-22
List widget with only 1 entry shall still show the drop-down


metas-mk:
### Is this a bug or feature request?
Feature Request
### What is the current behavior?
Currently, when pressing ```arrow-down``` on a focused list widget and the list has only 1 entry, then there is no drop-down list shown.
#### Which are the steps to reproduce?
Open, try and see.
### What is the expected or desired behavior?
If the drop-down list has only 1 entry, then show this one entry selected when opened.

metas-dh:
Results of IT1
tested in m15

* list field with only 1 entry shows the drop-down when pressing ```arrow-down```: OK
=> shows the 1 entry and "empty" where empty is optional, and only the 1 entry where it is not: OK

PR #1634, opened 2018-02-28
Show list dropdown for single options too

siemiatj:
Related to #1628 #1622

TODO:
* ~[ ] Create cypress test cases.~ On second thought I think I'll push tests as a separate PR.

### DIFF
--- a/src/components/widget/List/List.js
+++ b/src/components/widget/List/List.js
@@ -185,7 +185,7 @@ class ListWidget extends Component {
   activate = () => {
     const { list, listToggled } = this.state;
 
-    if (list && list.size > 1 && !listToggled) {
+    if (list && !listToggled) {
       this.setState({
         listToggled: true,
       });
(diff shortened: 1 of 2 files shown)


metas-mk:
Checked locally. List widgets now show drop-down when only 1 entry exists. OK
Lookup widgets now have issues described above. NOK.
@siemiatj pls ping me if demo is wanted

siemiatj:
@metas-mk Functional-wise should be good now.

metas-mk:
Quick local Test of
- single entry List widgets OK
- multi-entry List widgets OK
- Single entry Lookup widget OK (besides layout issue, creating follow-up)
- Multi entry Lookup widget OK (besides layout issue, creating follow-up)
- Location Editor OK
- Attributes Editor OK (besides that is does not close after opening. fix will follow in a few mins after merge)
--> Merging

Issue #1621, opened 2018-02-22
Dropdown List Widget - first entry at top shall be selected entry


metas-mk:
### Is this a bug or feature request?
Feature Request
### What is the current behavior?
Currently, when pressing ```arrow-down``` on a focused list widget, the drop-down list opens and the selected line is "somewhere".
#### Which are the steps to reproduce?
Open, try and see.
### What is the expected or desired behavior?
Opening a list widget drop-down list shall show
- the first item selected
- if the field already had a selected content, then the first item at the top shall be the selected item.

siemiatj:
Related #1550

siemiatj:
In case there's a default/selected value -> move it to the top

metas-ts:
Hi @siemiatj 
as talked: at least in process parameter dialogs, mandatory list fileds are not marked as mandatory.
to reproduce, you can go to 
https://w101.metasfresh.com:8443/window/540359 and do the "Create new procurement contract" process. All the list fields are actually mandatory, but they are not marked blue'ish (pls ignore the red arrow etc on the screenshot): 
![image pasted at 2018-3-1 13-00](https://user-images.githubusercontent.com/1264789/36892661-bb9df35a-1e05-11e8-9850-31aa24811b18.png)
That means that an infrequent user of that process might press start and get an error message four times (fo each one of the list fields that have no default value

)

metas-rc:
Results of  IT


Tested on webui


* OK: When I open a list, the first item on the list is selected
* OK: If there is a default item, that item is initially selected and displayed at the top of the list
* OK: If there are no default items and there is a none option, the none option is selected and at the top of the list
* OK: If there is no default item and no none option, the fist item (in alphanumeric order) is selected and displayed at the top of the list.

PR #1643, opened 2018-03-01
Selected entry at the top of dropdown select list

siemiatj:
When opening select dropdown the currently selected (or default) option should be at the top.

**Created a tmp branch as it's branched off of another branch which is not yet merged.**
Needs merging #1633 first and then a rebase with updated master. Other than that it's ready to test.

@metas-ts please check if this fixes your problem with mandatory field.

Related to #1621 and #1550


metas-mk:
@siemiatj All stuff above can be handled in follow-up issues if wanted. Pls, decide if to adjust directly or later.

PR #1647, opened 2018-03-02
Mandatory list widget border

siemiatj:
Add a blue border when field is mandatory, but leave the bottom dark line indicating focused state.


metas-mk:
Quick check locally. Works nicely. Merging.

Issue #1605, opened 2018-02-16
Scrolling in dropdown fields


metas-lc:
### Is this a bug or feature request?
bug
### What is the current behavior?
scrolling with mouse click (dragging the scrollbar) and with keyboard's arrows is not working well
#### Which are the steps to reproduce?
1. in this window we have lots of data in dropdowns so it's easier
https://w101.metasfresh.com:8443/window/540173/1000002
2. select the dropdown of Document Base Type field
3. use down key => notice that the scrollbar remains in it's place.  NOK
4. try to select the dropdown with the mouse click => the whole dropdown is gone NOK
### What is the expected or desired behavior?
step 3 Shall go down with your current selections
step 4 scrollbar shall move as you drag it. Shall not close

PabloSichert:
Part of this issue (arrow keys) is covered in https://github.com/metasfresh/metasfresh-webui-frontend/issues/1544

metas-lc:
IT
repeated the testcase above
3. goes down with my current selections OK
4. scrollbar moves as i drag it OK

Issue #1544, opened 2018-01-31
List Widget does not scroll down with arrow-down


metas-mk:
### Is this a bug or feature request?
Bug
### What is the current behavior?
Currently, the List widget does not scroll down when using the arrow-down. 
#### Which are the steps to reproduce?
- Open window/540321
- Open Filter
- Select Country and try to scroll down with arrow-down, to the bottom of the list.
### What is the expected or desired behavior?
Shall work.

![image](https://user-images.githubusercontent.com/13365687/35608375-296881bc-0659-11e8-85b9-6bef8513a9ca.png)

metas-rc:
Results of IT

Tested on webui



* OK: The dropdown list works well with up and down arrows.
* NOK: If I click on the dropdown and the mouse cursor stays somewhere in that list, I can use the arrow down and the entries are highlighted as selected one by one. But  if I stop pressing the arrow, the highlighted item will be the one from where the cursor is. Followup created: https://github.com/metasfresh/metasfresh-webui-frontend/issues/1666

PR #1652, opened 2018-03-05
Keep active item in view when scrolling list by arrow keys

PabloSichert:
#1544 and #1605 since they are closely related

### DIFF
--- a/src/components/widget/List/RawList.js
+++ b/src/components/widget/List/RawList.js
@@ -7,6 +7,9 @@ import classnames from 'classnames';
 
 import RawListDropdown from './RawListDropdown';
 
+const UP = Symbol('up');
+const DOWN = Symbol('down');
+
 const setSelectedValue = function(dropdownList, selected) {
   const changedValues = {};
   let idx = 0;
@@ -34,6 +37,7 @@ class RawList extends PureComponent {
 
     this.state = {
       selected: props.selected || null,
+      direction: null,
       dropdownList: props.list,
     };
   }
@@ -111,8 +115,42 @@ class RawList extends PureComponent {
       );
     }
 
-    this.checkIfDropDownListOutOfFilter();
+    // Delaying with requestAnimationFrame is needed to get the latest ref
+    requestAnimationFrame(() => {
+      this.checkIfDropDownListOutOfFilter();
+      this.scrollIntoView();
+    });
   }
+
+  scrollIntoView = () => {
+    const { list, selected } = this;
+
+    if (!list || !selected) {
+      return;
+    }
+
+    const { direction } = this.state;
+
+    const { top: topMax, bottom: bottomMax } = list.getBoundingClientRect();
+
+    const { top, bottom } = selected.getBoundingClientRect();
+
+    switch (direction) {
+      case UP:
+        if (top < topMax) {
+          selected.scrollIntoView(true);
+        }
+
+        break;
+
+      case DOWN:
+        if (bottom > bottomMax) {
+          selected.scrollIntoView(false);
+        }
+
+        break;
+    }
+  };
 
   checkIfDropDownListOutOfFilter = () => {
     if (!this.list) {
@@ -199,14 +237,14 @@ class RawList extends PureComponent {
       switch (e.key) {
         case 'ArrowUp':
           e.preventDefault();
-          this.navigate(false);
+          this.navigate(true);
           break;
         case 'ArrowDown':
           e.preventDefault();
           if (!isToggled) {
             onOpenDropdown();
           } else {
-            this.navigate(true);
+            this.navigate(false);
           }
           break;
         case 'Enter':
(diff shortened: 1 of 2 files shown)


siemiatj:
I rebased with master and did a quick refactor as some of the previously fixed issues resurfaced again.

One thing I noticed was when user opens the dropdown list and scrolls with keys, but mouse is in focus. It makes the scrolling jerky.

jscr47waot

src/components/widget/List/RawList.js
```diff
@@ -527,4 +584,4 @@ RawList.propTypes = {
   onCloseDropdown: PropTypes.func.isRequired,
 };
 
-export default onClickOutside(RawList);
```
siemiatj: Removing this broke the lookup list dropdown hide/show functionality on clicking outside.

src/components/widget/List/RawList.js
```diff
@@ -146,46 +174,61 @@ class RawList extends PureComponent {
...
+    const { list, handleCloseDropdown } = this;
+    const { isToggled } = this.props;
+
+    if (!list || !isToggled) {
+      return;
+    }
+
+    const { top, right, bottom, left } = list.getBoundingClientRect();
+
+    if (
```
siemiatj: I was thinking maybe we can use this instead : https://github.com/Pomax/react-onclickoutside#marking-elements-as-skip-over-this-one-during-the-event-loop ? 

But I'm not 100% sure we even need this.

src/components/widget/List/RawList.js
```diff
@@ -122,6 +121,35 @@ class RawList extends PureComponent {
     });
   }
 
+  handleOpenDropdown = () => {
+    const { handleClickOutside } = this;
+    const { onOpenDropdown, onFocus } = this.props;
+
+    onOpenDropdown();
+    onFocus();
```
siemiatj: Order is important. We first focus the field, and then open the dropdown. This way it's easier to keep track of what's going on.

src/components/widget/List/RawList.js
PabloSichert: I removed it because it didn't work as intended and implemented it myself. With `react-onclickoutside` you can't click on the scrollbar. I'll roll back the changes and post a screencast

src/components/widget/List/RawList.js
```diff
@@ -326,8 +399,9 @@ class RawList extends PureComponent {
             'input-dropdown-list-option-key-on': selectedRow,
           }
         )}
-        onMouseEnter={() => this.handleSwitch(option)}
         onClick={() => this.handleSelect(option)}
+        onMouseEnter={() => this.handleSwitch(option)}
```
siemiatj: This has to stay to properly indicate the selected element. We want the selected value to have an indication on the dropdown, but move to the hovered element when mouse is used.

src/components/widget/List/RawList.js
PabloSichert: Basically this issue https://github.com/metasfresh/metasfresh-webui-frontend/issues/1605. Could be that your suggestion https://github.com/metasfresh/metasfresh-webui-frontend/pull/1652#discussion_r172499024 might fix that

src/components/widget/List/RawList.js
PabloSichert: Okay, but it needs to be implemented differently then. If you mark the row as selected by simply hovering, the scroll-into-view logic won't work. It will interfere when you are using the arrow keys while the mouse is still in the box

siemiatj: Yup. I think using keys should temporarily disable mouse events, until another mouse event is detected.

So it should work like : 
```
user hovers with mouse -> list scrolls ->
user clicks arrow -> disable mouse select (as user is clearly not using it now) ->  list scrolls 
```

some pseudocode : 

```
  navigate = up => {
    (...)
    this.setState({
      disableMouse: true,
      (...)
    });
  };

(...)
        onMouseEnter={() => {
          if (!this.state.disableMouse) {
            this.handleSwitch(option);
          }}}
```

siemiatj:
The idea behind my recent changes to list elements was to have two states of the element : focused, and toggled. They're both controlled by the state of the parent List component. This way it's easier to keep track of what's going on there. It's still not perfect, but at least we don't have focused/blurred events on each level flying around. It's important to keep this order : element is first focused, and then becomes active. When using keyboard we have two separate states. When using mouse, it should toggle through the focused state automatically.

src/components/widget/List/RawList.js
siemiatj: What about https://github.com/Pomax/react-onclickoutside#regulate-whether-or-not-to-listen-to-scrollbar-clicks ?

PabloSichert:
Closing this so that #1544 and #1605 will be separated again

PabloSichert:
@siemiatj Next time, please don't rebase any branch before letting the author know, it will mess up the local checkout. It's fine if you put in some comments in the review feature and then we'll see what we do

src/components/widget/List/RawList.js
PabloSichert: That didn't seem to work, unfortunately

src/components/widget/List/RawList.js
PabloSichert: Included in https://github.com/metasfresh/metasfresh-webui-frontend/commit/146550d663515bd7d1446d18b07f798609dc54e4

Issue #1640, opened 2018-03-01
Drag and Drop Attachments


metas-lc:
### Is this a bug or feature request?
fr
### What is the current behavior?
right now you can't drag and drop multiple files at the same time
#### Which are the steps to reproduce?
1. open e.g. sales order
2. drag and drop 2 files
=> one uploaded, one not
![image](https://user-images.githubusercontent.com/16001666/36842887-d261713a-1d55-11e8-91dd-35f22730b560.png)

### What is the expected or desired behavior?
should be able to upload multiple files at the same time

PabloSichert:
The following translation keys need to be added:

- `window.attachment.title.pending` "Files pending"
- `window.attachment.title.done` "Files done"
- `window.attachment.title.error` "Files error"

metas-lc:
IT
1. make a new order, save
2. drag and drop 3 files
=> only 1 or 2 files are uploaded NOK
=> error: 
org.postgresql.util.PSQLException: ERROR: duplicate key value violates unique constraint "ad_attachment_record_id_ad_table_id"
  Detail: Key (record_id, ad_table_id)=(1000009, 259) already exists.
	SQL: INSERT INTO AD_Attachment (AD_Attachment_ID,AD_Client_ID,AD_Org_ID,AD_Table_ID,Created,CreatedBy,IsActive,Record_ID,Title,Updated,UpdatedBy) VALUES (nextval('ad_attachment_seq'),1000000,1000000,259,TO_TIMESTAMP('2018-03-12 17:08:14','YYYY-MM-DD HH24:MI:SS'),2188223,'Y',1000009,'.',TO_TIMESTAMP('2018-03-12 17:08:14','YYYY-MM-DD HH24:MI:SS'),2188223) RETURNING AD_Attachment_ID

![image](https://user-images.githubusercontent.com/16001666/37295287-7a41f942-2620-11e8-97f7-e52e905d7b37.png)

2. remove the attachments and drag and drop them again
=> works: all 3 files are imported, without errors OK
(except the trl errors. Follow-up: https://github.com/metasfresh/metasfresh/issues/3664)

PabloSichert:
@teosarca is this a backend issue or does the client need to submit files differently?

siemiatj:
@teosarca from error messages looks like either we need some unique keys/id's for each file, or backend needs some changes.

PR #1664, opened 2018-03-08
Allow to drag / drop multiple files at once

PabloSichert:
#1640

### DIFF
--- a/src/components/Dropzone.js
+++ b/src/components/Dropzone.js
@@ -10,19 +10,19 @@ class DropzoneWrapper extends Component {
     };
   }
 
-  handleDropFile(accepted, rejected) {
+  handleDropFile = (accepted, rejected) => {
     const { handleDropFile, handleRejectDropped } = this.props;
 
     this.handleDragEnd();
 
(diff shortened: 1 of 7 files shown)


src/components/Dropzone.js
```diff
@@ -10,19 +10,19 @@ class DropzoneWrapper extends Component {
...
 
-  handleDropFile(accepted, rejected) {
+  handleDropFile = (accepted, rejected) => {
     const { handleDropFile, handleRejectDropped } = this.props;
 
     this.handleDragEnd();
 
-    if (accepted.length) {
-      handleDropFile(accepted);
+    for (const file of accepted) {
```
siemiatj: I'm a fan of using functional style and would go with map instead. For-of is a huge overhead for a simple array.

PabloSichert: `map` returns an array that we are not using here, so `forEach` would reveal the intention a bit more.

Personally I think the `for-of` makes for really readable code, but I'm fine using either

metas-mk:
checked locally. Works as required. Merged.

PR #1665, opened 2018-03-09
Don't show change in all rows

siemiatj:
Small but deadly typo.


metas-rc:
Note: This issue was about modifying a value in a grid row (e.g. Sales Order Line's qty)  and it happened that all the rows were updated to that qty. The change did not persist after refreshing.

Results of IT

Tested on webui

* OK: If I change a qty in an order line, the qtys in other lines are not altered.

Issue #1669, opened 2018-03-12
Empty option missing in dropdown


metas-lc:
### Is this a bug or feature request?
bug
### What is the current behavior?
if you have a dropdown containing "none", firt time that you want to change it to none, the option is missing. Second time is visible
#### Which are the steps to reproduce?
1. go to business partner and select one. E.g. https://w101.metasfresh.com:8443/window/123/2156441
2. go to customer tab, check price system
![image](https://user-images.githubusercontent.com/16001666/37285803-dd8076ae-2608-11e8-9041-03c29c573742.png)

3. set it on some price system, check again
=> none is missing
4. try to change it again
=> none is there

(you can also reproduce it in other places, like HU Editor: label attribute)
### What is the expected or desired behavior?
should be able to see "none" each time you open the dropdown

metas-lc:
IT
retested the testcase above => works fine OK

PR #1671, opened 2018-03-12
Fix none option disappearing from list options

siemiatj:
It was foolish to use data from props instead of local copy which contains the default option as well.

Added basic cypress e2e test for list widget in table, and basic automated test. To be able to write proper automated test we first need to figure out how to render Tethered components with enzyme.

Relates to #1669

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,5 +15,8 @@ dist
 # Coverage report
 .nyc_output
 coverage
+test_setup
+src/__tests__
+cypress
 
 package-lock.json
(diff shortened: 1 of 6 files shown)


PabloSichert:
> To be able to write proper automated test we first need to figure out how to render Tethered components with enzyme.

Unit test could (and should be) written independently from what functionality `react-tether` provides.

That means that either
- only when exporting the component is wrapped in a tether or
- we mock out tether by just returning the children

siemiatj:
Problem with Tethered components is not the fact that they're wrapped, but that they're attached to the window. I haven't really looked into how we can get around that so I'm open to any suggestions.

PR #1682, opened 2018-03-16
Improve UI for smaller devices

siemiatj:
Bunch of improvements around tables and modals for smaller factor devices. Removed pagination in tables (instead we're requesting the full list), hid columns if a corresponding param is set and decreased margins/paddings. Using `current-device` package to decide what device is being used, and set css styles accordingly.
Tested on Pixel XL Emulator with 2560x1600px resolution.

Additionally I've added a button for included views, that can change the width of panels. @metas-mk  let me know if we want to keep it, or not : 

efkalt1rdq

Related to #1673


src/utils/index.js
```diff
@@ -0,0 +1,16 @@
+import queryString from 'query-string';
+
+export const getQueryString = query =>
+  queryString.stringify(
+    Object.keys(query).reduce((parameters, key) => {
```
PabloSichert: You can use `Object.entries` to iterate over key/value pairs

Issue #1684, opened 2018-03-18
Combine dropdown list of <List> and <Lookup> component


PabloSichert:
### Is this a bug or feature request?
Issue combining multiple bugs: #1544, #1605, #1641

### What is the current behavior?
Both `<List>` and `<Lookup>` widgets implement their own dropdown lists. Behaviour is not consistent and both of them have their own set of bugs.

### What is the expected or desired behavior?
Merge functionality by sharing the same dropdown component and have it work as intended

metas-rc:
Results of IT


* #1641 IT OK
* #1544 IT OK
* #1605 IT OK - exception: The scrollbar is still not working but it will be covered in #1606

PR #1685, opened 2018-03-18
Refactor and merge dropdown lists of <List> and <Lookup>

PabloSichert:
#1684

`RawList` and `RawLookup` are still a bit of a tangled mess. I kept them like this for now and refactored their dropdown lists.

@metas-mk are there Lookup components in the app where new items can be added? If so, could you give me a link? (I might have removed that functionality and need to have a closer look)


src/components/widget/List/List.js
```diff
@@ -60,11 +60,11 @@ class ListWidget extends Component {
     if (prevProps.autoFocus !== autoFocus && !isFocused) {
       if (autoFocus) {
         this.handleFocus();
-        !doNotOpenOnFocus && list.size > 1 && this.activate();
+        !doNotOpenOnFocus && list && list.size > 1 && this.activate();
```
siemiatj: Is it possible for list to not exist ? Isn't it easier to just always create an empty List object, and have one less point of failure ?

src/components/widget/Lookup/RawLookup.js
```diff
@@ -407,51 +327,43 @@ class RawLookup extends Component {
...
-              to: 'scrollParent',
-            },
-            {
-              to: 'window',
-              pin: ['bottom'],
-            },
-          ]}
+        <div
+          className={
+            'raw-lookup-wrapper raw-lookup-wrapper-bcg' +
```
siemiatj: use `classnames` ?

src/components/widget/List/List.js
PabloSichert: I didn't investigate much into this (the reason being that it needs refactoring anyway.)

At some point I got an exception since list was undefined, so I threw that hotfix in there

siemiatj:
Looks good overall but needs thorough testing to be sure we're not breaking the current functionality.

PabloSichert:
Agree.

I'd say we should be able to iron out potential bugs until Friday

PabloSichert:
Adding new item implemented

metas-mk:
Checked. New works now. Merging.

Issue #1681, opened 2018-03-15
Data Entry via barcode scan in Action Parms


metas-mk:
### Is this a bug or feature request?
Feature Request
### What is the current behavior?
Currently, a new feature was built that allows the barcode scanning via webcam/ camera. This does not work in action parms/ process parameters.
#### Which are the steps to reproduce?
Open, try and see.
- Window https://w101.metasfresh.com:8443/window/540373
- Action Move HU
- Allow the scan of "Locator To"
### What is the expected or desired behavior?
Possibility to trigger a barcode scan via camera when focussed in a parameter.

![image](https://user-images.githubusercontent.com/13365687/37473609-cbc884d4-286e-11e8-9433-8be83054baaa.png)

metas-dh:
Results of IT1
tested in mf15

Possibility to trigger a barcode scan via camera when focussed in a parameter:
* barcode scan worked fine (used barcode from SSCC label): OK
* i still had to set the HU manually, and could select HUs which were not the ones my barcode was taken from: NOK i think, but not sure if this is relevant in this issue? @metas-mk wdyt?
* i could stop the camera with using ```X```, but the modal remained, i had to refresh the page to leave: NOK imho

siemiatj:
I don't have real barcodes to test this, and there was nothing about automatically selecting HU's or limiting the options in the ticket.

For the last point, I'll have a PR in a few.

metas-dh:
Results of IT2
tested in mf15:

 * stopping the camera by pressing the X worked fine: OK

metas-dh:
>i still had to set the HU manually, and could select HUs which were not the ones my barcode was taken from: NOK i think, but not sure if this is relevant in this issue? 

checked w/ metas-mk: OK, HU selection will be done before scanning

PR #1690, opened 2018-03-20
Barcode scanner in lookup widgets

siemiatj:
Rewrite BarcodeScanner so that it's easy to add it to other fields.

Relates to #1681


src/assets/css/inputs.css
```diff
@@ -659,35 +659,43 @@ input:checked + .input-slider:before {
...
+        }
+
+        &&.focused {
+            border-bottom: 1px solid #4b5b64;
+        }
+
+        &&.opened {
+            border-bottom: none;
+        }
+    }
```
PabloSichert: Nice refactoring! 

Small heads-up: This will clash with https://github.com/metasfresh/metasfresh-webui-frontend/pull/1687/commits/2dee0652ea1336da44667e894f62d9d6ad50336c #1687

src/assets/css/inputs.css
```diff
@@ -659,35 +659,43 @@ input:checked + .input-slider:before {
     cursor: pointer;
     min-width: 200px;
     padding: .125rem .5rem;
+
+    &&:focus {
```
PabloSichert: Small nitpick: Does `&&` do anything different than `&`?

Could find any reference in https://sass-lang.com/documentation/file.SASS_REFERENCE.html#parent-selector

src/assets/css/inputs.css
```diff
@@ -819,26 +818,48 @@ input:checked + .input-slider:before {
...
+    .input-dropdown-list-option:hover * {
+        color: $brand-font-color;
+        background-color: $brand-danger-color !important;
+    }
+
+    .input-dropdown-list-option.input-dropdown-list-option-key-on,
+    .input-dropdown-list-option.input-dropdown-list-option-key-on * {
+        color: $brand-font-color;
+        background-color: $brand-danger-color !important;
+    }
```
PabloSichert: While you're at it, might as well use `&` to nest those

src/components/widget/BarcodeScanner/BarcodeScannerWidget.js
```diff
@@ -0,0 +1,77 @@
...
+              result={result}
+              onDetected={this._onBarcodeDetected}
+              onClose={() => this._scanBarcode(false)}
+              onReset={() => this._scanBarcode(true)}
+            />
+            <div className="row scanning-result">
+              <span className="label col-xs-3 col-sm-2">Barcode</span>
+              <BarcodeScannerResult
+                result={result}
+                onSelect={result => this.selectBarcode(result)}
```
PabloSichert: You can change this line to

```js
onSelect={this.selectBarcode}
```

and potentially save some re-renders of `BarcodeScannerResult` by not creating a new function object every time

src/components/widget/Lookup/Lookup.js
```diff
@@ -249,6 +280,19 @@ class Lookup extends Component {
     const errorInputCondition =
       validStatus && (!validStatus.valid && !validStatus.initialValue);
     const classRank = rank || 'primary';
+    let showBarcodeScannerBtn = false;
+
+    if (scanning) {
+      return (
+        <div
+          className="overlay-field js-not-unselect"
+          onKeyDown={e => e}
```
PabloSichert: That looks like you forgot to remove 😄

src/components/widget/BarcodeScanner/BarcodeScannerWidget.js
```diff
@@ -0,0 +1,77 @@
...
+  return class BarcodeScannerWidget extends Component {
+    constructor(props) {
+      super(props);
+
+      this.state = {
+        scanning: false,
+        result: null,
+        barcodeSelected: null,
+      };
+    }
```
PabloSichert: You can remove the constructor here and declare it as an instance variable, since it doesn't depend on props

```js
state = {
  // ...
};
```

src/components/widget/BarcodeScanner/BarcodeScannerWidget.js
```diff
@@ -0,0 +1,77 @@
...
+
+    _scanBarcode = val => {
```
PabloSichert: Does the leading underscore have any implication here or could it be removed?

src/assets/css/inputs.css
siemiatj: Oh, I don't know to be honest. I've been using it for years and seems to be working. But you're right, it's not really mentioned anywhere.

Just for the record, we're using postcss and whatever it supports.

src/assets/css/inputs.css
siemiatj: Yeah the thing is I wasn't really focusing on cleaning up css here. And I'm reluctant to rewrite too much at once as it's easy to break things. So I'm just gonna keep fixing those things gradualy while I work on other stuff.

src/components/widget/BarcodeScanner/BarcodeScannerWidget.js
siemiatj: Nice, thanks for catching this.

src/components/widget/Lookup/Lookup.js
siemiatj: Yeah, I'm not really sure what I was thinking here.

src/components/widget/BarcodeScanner/BarcodeScannerWidget.js
siemiatj: True, it's not needed in the HOC anymore.

src/components/widget/BarcodeScanner/BarcodeScannerWidget.js
siemiatj: It's a leftover from when it was not a standalone HOC.

src/components/widget/BarcodeScanner/BarcodeScannerWidget.js
PabloSichert: Not sure I understand - why? The state won't be shared between instances. It will be created again for each new instance (as would happen if you create it in the constructor)

src/assets/css/inputs.css
PabloSichert: Good point.

By the way, we should get source maps for CSS working. Otherwise it's really hard to find those composite class names

src/components/widget/BarcodeScanner/BarcodeScannerWidget.js
PabloSichert: You still need to call `setState` instead of mutating `this.state` directly, of course

siemiatj: Changing instance variable doesn't trigger render, so the wrapped component won't be re-rendered with the new value in props. So in this case if I'd change the value of `scanning` (which controls if scanner widget is visible or not), widget would stay opened.

PabloSichert: See my comment above, you still need to call `setState`. The comment was only concerning the instantiation of `this.state`. (Doing `this.state = ...;` in the constructor and declaring an instance variable `state = ...;` are equivalent)

siemiatj: No, they're not :) Changing instance variables does not trigger render, as they're utside of React flow. 

From:
https://www.reddit.com/r/reactjs/comments/54gl72/why_use_state_vs_class_properties/

```
Setting a field directly on the component instance, like this.someField = 42, will not cause the component to actually re-render.
```

PR #1691, opened 2018-03-20
Open time widgets with time view (instead of date view)

PabloSichert:
#1654


src/components/widget/TetheredDateTime.js
```diff
@@ -1,9 +1,24 @@
 import React from 'react';
+import PropTypes from 'prop-types';
 import DateTime from 'react-datetime';
 import CalendarContainer from 'react-datetime/src/CalendarContainer';
 import TetherComponent from 'react-tether';
 
 export default class TetheredDateTime extends DateTime {
+  static propTypes = {
```
siemiatj: Can we stick to how propTypes are defined in all the code examples as well as the `prop-types` documentation ? Static properties are still an experimental feature so we should avoid it.

https://www.npmjs.com/package/prop-types

PabloSichert: Oh, you are right - I assumed they were finalized by now. However, they are [Stage 3](https://github.com/tc39/proposal-class-fields) already and also mentioned in the [prop type docs (last code block)](https://reactjs.org/docs/typechecking-with-proptypes.html#default-prop-values).

When they are declared at the top of the class they are more coupled to the class and not that easy to overlook, in my opinion.

What do you think?

PabloSichert:
Interesting, I need to investigate how the backend stores time only. It looks like it still expects a full date

src/components/widget/TetheredDateTime.js
siemiatj: But then again propTypes are used by the redux HOC wrapper which is at the bottom. Also mind that we should start adding some code comments and documentation. What would happen to the class code if we had those props (some old code from a different project):

```javascript
/**
 * PropTypes.
 *
 * @property {object} project - project data object
 * @property {object} revisions - object with all revisions
 * @property {array} comments - all entry comments
 * @property {object} entryState - object with single entry state
 * @property {object} user - user data object
 * @property {array} visibleEntries - list of all visible entries for current gallery page
 * @property {object} entry - entry data object
 * @property {object} revision - revision id
 * @property {object} allEntries - object with all entries in the state by id
 * @property {number} itemsCount - total number of elements
 * @property {number} pageSize - number of elements per page
 * @property {number} currentPage - current gallery page number
 * @property {func} fetchComments - action creator for getting entry comments
 * @property {func} scoreRevision - action creator for scoring a revision
 * @property {func} zoomInEntry - action creator to zoom-in the image
 * @property {func} zoomOutEntry - action creator to zoom-out the image
 * @property {func} resetZoom - action creator to reset zoom to 1
 * @property {func} deleteEntry - action creator to remove entry (mark it accordingly)
 * @property {func} fetchProjectEntry - action creator for fetching a single entry by id
 * @property {func} setEntryDetailsPage - action creator for saving the current entry details
                    page number in the state
 * @property {func} toggleEntryDetails - show/hide single entry modal
 */
EntryDetails.propTypes = {
  project: PropTypes.object,
  revisions: PropTypes.object,
  comments: PropTypes.object,
  entryState: PropTypes.object,
  user: PropTypes.object,
  visibleEntries: PropTypes.array,
  entry: PropTypes.object,
  revision: PropTypes.number,
  allEntries: PropTypes.object,
  itemsCount: PropTypes.number,
  pageSize: PropTypes.number,
  currentPage: PropTypes.number,
  fetchComments: PropTypes.func,
  scoreRevision: PropTypes.func,
  zoomInEntry: PropTypes.func,
  zoomOutEntry: PropTypes.func,
  resetZoom: PropTypes.func,
  deleteEntry: PropTypes.func,
  fetchProjectEntry: PropTypes.func,
  setEntryDetailsPage: PropTypes.func,
  toggleEntryDetails: PropTypes.func,
};
```

I'm not gonna force this or fight with my life to defend this approach, but it's just easier and cleaner for me :) I don't think it's worth spending more than 5s on that.

Issue #1675, opened 2018-03-14
Tab Sequence


metas-lc:
### Is this a bug or feature request?
fr
### What is the current behavior?
currently, when we are using tab in a document, the fields are not taken in order. It jumps over some fields
also, in modal window, if you get at the end of the fields, the tab is not starting again at the beginning.
#### Which are the steps to reproduce?
1. create a new sales order, add bpartner, press tab multiple times
=> notice that from "Diss. unloading point" jumps to OrderLine included tab. Shall go to Preissystem first
2. go to picking terminal window, select a line -> pick
3. select a slot, open hu selection window. Select an HU and use "Als Quell-HU ..."
4. select the picking slot again, dropdown - "Pick to new HU", use tab in there
=> if you are on qty and press tab, is not going back to first field (should go)
=> and now i saw that if you are on first field and use tab, it jumps back NOK

### What is the expected or desired behavior?
1. Shall go to Preissystem first
2. should go back to 1st tab
3. shall not jump back

metas-lc:
IT
1. sales order - pressing tab from bpartner field 
=> it takes all editable fields in order: header left and right, included tabs, add new, batch entry, docstatus OK
2. using tab in "pick to new HU"
=> if you are on qty and press tab, is not going back to first field (should go) NOK
=> if you are on first field and use tab, it doesn't jump back OK
3.

metas-lc:
IT
1. sales order - pressing tab from bpartner field
=> it takes all editable fields in order: header left and right, included tabs, add new, batch entry, docstatus OK
note: if you keep pressing tab, and fields switch really fast, some fields are set on null. We get errors because of the mandatory ones. Created tasks for this
https://github.com/metasfresh/metasfresh-webui-frontend/issues/1810
https://github.com/metasfresh/metasfresh-webui-api/issues/970
(to see if i have to create one for not setting fields on null using tab)
2. using tab in "pick to new HU"
=> if you are on qty and press tab, is going back to first field OK
=> if you are on first field and use tab, it doesn't jump back OK

PR #1695, opened 2018-03-21
Fix tab sequence for form fields

PabloSichert:
#1675 

- disabled widgets are now skipped
- widgets behind modals don't steal focus any more
- inbox now opens when reaching with tab

foo

bar

### DIFF
--- a/src/components/filters/FiltersFrequent.js
+++ b/src/components/filters/FiltersFrequent.js
@@ -4,7 +4,7 @@ import React, { Component, Fragment } from 'react';
 import { connect } from 'react-redux';
 import onClickOutside from 'react-onclickoutside';
 
-import { TableCell } from '../table/TableCell';
+import TableCell from '../table/TableCell';
 import FiltersDateStepper from './FiltersDateStepper';
 import FiltersItem from './FiltersItem';
 
@@ -45,6 +45,7 @@ class FiltersFrequent extends Component {
       applyFilters,
       clearFilters,
       active,
+      modalVisible,
     } = this.props;
 
     const { openFilterId } = this.state;
@@ -80,6 +81,7 @@ class FiltersFrequent extends Component {
                   ['btn-active']: item.isActive,
                   ['btn-distance']: !dateStepper,
                 })}
+                tabIndex={modalVisible ? -1 : 0}
               >
                 <i className="meta-icon-preview" />
                 {item.isActive &&
@@ -134,6 +136,7 @@ class FiltersFrequent extends Component {
 
 const mapStateToProps = state => ({
   allowOutsideClick: state.windowHandler.allowOutsideClick,
+  modalVisible: state.windowHandler.modal.visible,
 });
 
 export default connect(mapStateToProps)(onClickOutside(FiltersFrequent));
--- a/src/components/inbox/Inbox.js
+++ b/src/components/inbox/Inbox.js
@@ -91,13 +91,14 @@ class Inbox extends Component {
   };
 
   render() {
-    const { open, inbox, all, close } = this.props;
+    const { open, inbox, all, close, modalVisible } = this.props;
+
     return (
       <div
         className="js-inbox-wrapper"
         onKeyDown={e => this.handleKeyDown(e)}
         onFocus={this.props.onFocus}
-        tabIndex={0}
+        tabIndex={modalVisible ? -1 : 0}
       >
         {(all || open) && (
           <div className={all ? 'inbox-all ' : 'inbox'}>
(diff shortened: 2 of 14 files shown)


src/components/filters/FiltersFrequent.js
```diff
@@ -140,6 +141,7 @@ class FiltersFrequent extends Component {
 
 const mapStateToProps = state => ({
```
siemiatj: We could get `({ windowHandler })` directly.

src/components/inbox/Inbox.js
```diff
@@ -66,14 +66,6 @@ class Inbox extends Component {
     deleteUserNotification(item.id).then(() => {});
   };
 
-  componentDidUpdate() {
```
siemiatj: 👍 for removing as many lifecycle methods as possible

src/components/inbox/Inbox.js
```diff
@@ -94,12 +91,14 @@ class Inbox extends Component {
...
   render() {
-    const { open, inbox, all, close } = this.props;
+    const { open, inbox, all, close, modalVisible } = this.props;
+
     return (
       <div
         className="js-inbox-wrapper"
         onKeyDown={e => this.handleKeyDown(e)}
-        tabIndex={0}
+        onFocus={this.props.onFocus}
```
siemiatj: Why not get it from props ? ^^

src/components/table/TableCell.js
```diff
@@ -218,4 +220,6 @@ class TableCell extends PureComponent {
   }
 }
 
-export default TableCell;
+export default connect(state => ({
+  modalVisible: state.windowHandler.modal.visible,
```
siemiatj: Can't we grab this in one of the parent containers that is already connected to the store ? We should limit the amount of components subscribing to the store and create containers that handle data fetching instead.

src/components/filters/FiltersFrequent.js
PabloSichert: Yeah, that variable should be shared.

```js
const mapStateToProps = state => {
  const { allowOutSideClick, modalVisible } = state.windowHandler;

  return {
    allowOutSideClick,
    modalVisible,
  };
};
```

I think this reveals better what comes from where (especially "state" is named and not anonymously destructured). What do you think?

src/components/inbox/Inbox.js
PabloSichert: Do you mean `tabIndex`? I just kept the `0` for now, since it's a sensible default and this component is only used in one place. We can add that prop later should we need it

src/components/table/TableCell.js
PabloSichert: Good catch, that piece of code is definitively worth talking about.

Since it's an interesting argument, I hope you don't mind if I elaborate a bit. Generally I like how `connect` makes a component self-contained. It works really well, especially if the usage of the component is always the same.

Consider the following example:

```js
// TableCell expects a "modalVisible" prop. It doesn't really tell a good story of why it needs it

ComponentA(props) {
  return <TableCell modalVisible={props.modalVisible} />;
}

ComponentB(props) {
  return <TableCell modalVisible={props.modalVisible} />;
}

ComponentC(props) {
  return <TableCell modalVisible={props.modalVisible} />;
}
```

vs

```js
// TableCell is connected to "modalVisible" and does not concern its parent components about it

ComponentA() {
  return <TableCell />;
}

ComponentB() {
  return <TableCell />;
}

ComponentC() {
  return <TableCell />;
}
```

Also performance-wise I'd argue that connecting `TableCell` is better. If `modalVisible` changes, only the cells will re-render, since ultimatively their render output is dependent on that value. If you connect parent components, they will re-render without having changed anything, only for the purpose to provide `TableCell` with a new value.

> https://github.com/reactjs/redux/blob/master/docs/faq/Performance.md
> [...] many individual components should be connected to the store instead of just a few, and connected list components should pass item IDs to their connected child list items (allowing the list items to look up their own data by ID). This minimizes the overall amount of rendering to be done.

---

On a side note: Regarding the container/component debate, you might be interested in reading this:

https://twitter.com/dan_abramov/status/802569801906475008

src/components/inbox/Inbox.js
PabloSichert: Whoopsie, haha, got it

src/components/table/TableCell.js
siemiatj: Well yeah, normally I would agree. In this case though : 
- it's just a single prop, so doesn't change much. Unless we'll define TableCell props as {...props} instead of defining them separately, it won't really matter.
- I agree that there's no reason to be anal about component/container definition and separation. The thing is though our components are really complicated and often 1000 lines long. So in this case, rewriting them into two files would help with maintainability and readability.

Last but not least, quoting Dan:
```
Components often flipped back and forth as we worked on them.
```

PabloSichert: I'm all about changing things as we go.

So if I understand correctly -  you want to wrap a component in a container that's doing the data fetching and then reusing that container instead of the component directly?

Yes, that would be perfectly valid to seperate the requirements of our larger components. We just need to care that not too much logic creeps into those.

For this component: I suppose you agree it doesn't make sense to introduce a container that fetches one prop - fine to keep it like that now?

siemiatj: Yup, for this component it makes no sense.

Issue #1649, opened 2018-03-02
Date range widget needs to support 2018-05-01 to 2019-04-30


metas-ts:
### Is this a bug or feature request?
Bug
### What is the current behavior?

Currently, I can't enter a date range where the day&month of the end date are smaller that the day&month of the start date - besides the year of the end date beeing bigger.

* I can't enter them using the calendar widget
* I can aparently enter them with the keyboard, like this:
![image](https://user-images.githubusercontent.com/1264789/36908150-bfc489d2-1e3a-11e8-8548-375e8624f8c6.png)

..but then then i press the "apply" button, the following is send to the API:
![image](https://user-images.githubusercontent.com/1264789/36908389-5e473f5a-1e3b-11e8-8471-849276e8c6e2.png)

.. i.e. 
* the datefrom is altered from what i entered
* as a user, i can't even see it in the widget

### What is the expected or desired behavior?

pls ping @metas-mk if there are questions :-)

siemiatj:
Not perfect, but changing some configs it's much more intuitive now.

![kr6dwshuda](https://user-images.githubusercontent.com/513460/37718136-50170e88-2d22-11e8-8588-e01e59dbe46d.gif)

metas-rc:
Results of IT

Tested on webui: 


* OK: I was able to insert the range 2018-05-01 to 2019-04-30 by selecting the dates in the widget, using mouse click ( like shown in the dev test gif)

* OK : The range persisted also after I pressed apply and the filter was correctly applied.

* OK: It also works when manually inserting the range

PR #1698, opened 2018-03-21
Improve UX of the DateTime picker widget

siemiatj:
Related to #1649


siemiatj:
This looks like some timezone thing. Need to check this.

Issue #1700, opened 2018-03-22
attribute filter fields are shrinked 


metasnw:
### Is this a bug or feature request?
bug

### What is the current behavior?
attribute filter fields are shrinked look like:
![image](https://user-images.githubusercontent.com/16079655/37761685-a90fdb3a-2dba-11e8-9586-372965012c4a.png)

#### Which are the steps to reproduce?
found on 5.49RC and latest master

1. go to /window/123
1. click filter

### What is the expected or desired behavior?
frame should be so big that the text fits in

metas-lc:
IT
checked attribute filter - it's big and the text fits in

PR #1702, opened 2018-03-22
Fix size of Labels dropdown

siemiatj:
Related to #1700


siemiatj:
I think this Label dropdown showing inside the filters modal is a new thing, and not a regression/bug.

siemiatj:
Moved Label dropdown to a tethered component, and fixed the cursor position : 

screen shot 2018-03-26 at 7 19 32 pm

Issue #1679, opened 2018-03-14
Layout bug with resolution 1920x1200


metasnw:
### Is this a bug or feature request?
bug

### What is the current behavior?
when using  resolution 1920x1200 the field layout is smalle as more fields are displayed and the widget size is ignored
![image](https://user-images.githubusercontent.com/16079655/37408894-5b084332-279d-11e8-944f-e7e69fb928df.png)

e.g. /window/143/1000537

once you change something (e.g. double click on qty) in the line the layout switches and the product field gets to the same size as with resolution 1920x1080:
![image](https://user-images.githubusercontent.com/16079655/37409173-edc95d3c-279d-11e8-8f96-3eb2902c8b41.png)

metasnw:
also saw it on 5.47 with resolution 1920x1080

siemiatj:
What should I be looking for here ? Screenshot from latest master and 1920x1200 in dev tools responsive mode :
 
![screen shot 2018-03-20 at 1 16 57 am](https://user-images.githubusercontent.com/513460/37628993-74fc0790-2bdc-11e8-9d36-c1124737d519.png)

siemiatj:
For ITs : it's enough to have the browser window at least 1200px wide to test this.

metas-lc:
checked product field from order window in my browser's resolution, in 1920x1200 and 1920x1080
=> the field has the same size and doesn't change if you make something in the lines. OK
other thing i notice that it's probably related (even though it happens in any resolution) : open contract window, see No. field: it's small at beginning and then, if you e.g. sort fields, it gets bigger
https://w101.metasfresh.com:8443/window/540359?viewId=540359-ff00c02322c74bb780dc318931b27a05&page=1
shall i create a new task?

metasnw:
> open contract window, see No. field: it's small at beginning and then, if you e.g. sort fields, it gets bigger

that is exactly the behavior I would have expected to be fixed in this issue

siemiatj:
In order to completely get rid of that, we need to set columns to a fixed widths.

metas-lc:
IT
1. the field still gets a bit larger when you use zoom in (110%)- zoom out (100%), but is not that visible and the product;s name fits OK?
2. checked No. field from contracts window => big enough, not affected by sorting OK

PR #1704, opened 2018-03-22
Fix widths of table columns

siemiatj:
Header `th` elements also need to have widths for their classes defined. Relates to #1679 

Huge cleanup of table-specific styles. @PabloSichert please take a look if I haven't missed anything obvious :)

### DIFF
--- a/src/assets/css/table.css
+++ b/src/assets/css/table.css
@@ -1,119 +1,183 @@
-
 .table {
     margin-bottom: 1px;
+
+     *::selection {
+        background-color: transparent;
+        color: $brand-font-color;
+    }
+
+    .input-body-container *::selection {
+        background-color: $brand-color-primary;
+        color: #fff;
+    }
+
+    &.table-included {
+        background-color: transparent;
+        height: 100%;
+        
+        td {
+            height: 22px;
+            width: 50%;
+        }
+    }
+
+    &.table-bordered-horizontally td {
+        border-bottom: 1px solid $brand-border-color;
+    }
+
+    &.table-bordered-vertically td:not(:last-child) {
+        border-right: 1px solid $brand-border-color;
+    }
+
+    &.table-bordered-vertically .item-caption td {
+        border-right: none;
+    }
+
+    &.table-bordered-vertically td.indented {
+        line-height: 28px;
+        min-width:0;
+        border-right:0;
+        padding-top:0;
+        padding-bottom:0;
+        vertical-align: middle;
+    }
+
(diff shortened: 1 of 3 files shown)


src/assets/css/table.css
```diff
@@ -1,119 +1,183 @@
...
+        }
+    }
+
+    /* Table cols sizing */
+    .td-lg {
+        width: 225px;
+        max-width: 225px;
+        min-width: 225px;
+
+        .input-dropdown-list {
```
siemiatj: @PabloSichert you recently removed the rule that was setting this width to 225px, but there was also another one that set it to auto. Does this collide with your changes ?

src/components/table/Table.js
```diff
@@ -953,87 +953,86 @@ class Table extends Component {
...
+            indentSupported,
+            selected,
+            docId,
+            tabIndex,
+            readonly,
+            collapsible,
+            viewId,
+          }}
+          key={`${i}-${docId}`}
+          collapsed={collapsedParentsRows.indexOf(item[keyProperty]) > -1}
```
PabloSichert: For readability, you could replace all

```js
···.indexOf(···) > -1
```

with

```js
···.includes(···)
```

src/assets/css/table.css
PabloSichert: I didn't test for functionality there. Just randomly discovered and removed an obsolete rule that would never apply. It was something like

```css
.x .y {
  width: 123px;
}

.x .y {
  width: auto;
}
```

src/assets/css/table.css
```diff
@@ -1,119 +1,183 @@
...
+    &.table-bordered-vertically .item-caption td {
+        border-right: none;
+    }
 
-.table.table-bordered-horizontally td {
-    border-bottom: 1px solid $brand-border-color;
-}
+    &.table-bordered-vertically td.indented {
+        line-height: 28px;
+        min-width:0;
```
PabloSichert: Just a small nitpick, there are some spaces missing here. You can leave them like this until we introduce a tool to automatically fix CSS formatting

src/assets/css/table.css
siemiatj: Yeah, it was a dupe. Ok.

siemiatj:
No idea what is Codacy trying to do here, but none of this errors that it's showing is an error.

metas-mk:
Checked locally. Fixed. Checked on 1980x1200/1080. Works. Merging.

PR #1711, opened 2018-03-26
Fix lists after last merge

siemiatj:
Bunch of different things stopped working after last refactoring. We need to be more diligent next time, as it should be easily noticeable. Also none of my comment from the refactor PR was ever addressed.

Fixed:
- closing the dropdown on mouse selection
- opening the dropdown with arrow down when the field is focused
- weird props error
- improved focused state indication

### DIFF
--- a/src/components/widget/SelectionDropdown.js
+++ b/src/components/widget/SelectionDropdown.js
@@ -1,6 +1,7 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import ReactCSSTransitionGroup from 'react-addons-css-transition-group';
+import classnames from 'classnames';
 
 export default class SelectionDropdown extends Component {
   static propTypes = {
@@ -95,7 +96,8 @@ export default class SelectionDropdown extends Component {
     this.ignoreNextMouseEnter = true;
 
     const { selected, options, onChange } = this.props;
-
+    const size = this.size(options);
+    const selectedNew = this.get(options, index);
     let index = options.indexOf(selected);
 
     if (up) {
@@ -104,15 +106,11 @@ export default class SelectionDropdown extends Component {
       index++;
     }
 
-    const size = this.size(options);
-
     if (index < 0) {
       index = 0;
     } else if (index >= size) {
       index = size - 1;
     }
-
-    const selectedNew = this.get(options, index);
 
     if (selectedNew === selected) {
       return;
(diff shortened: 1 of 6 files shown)


PabloSichert:
Which comments are you referring to? I can't remember anything being still open and it wasn't intentional otherwise

src/components/widget/SelectionDropdown.js
```diff
@@ -129,9 +130,11 @@ export default class SelectionDropdown extends Component {
 
     switch (event.key) {
       case 'ArrowUp':
+        event.preventDefault();
         navigate(true);
         break;
       case 'ArrowDown':
+        event.preventDefault();
```
PabloSichert: `preventDefault` is already called [here](https://github.com/metasfresh/metasfresh-webui-frontend/blob/master/src/components/widget/SelectionDropdown.js#L153)

siemiatj:
Yeah, I guess it's an overlook on all sides. I left some comments on the List/Lookup PR as far as I remember. And didn't do a great job reviewing this. Removing onOutsideClick handler was an obvious thing and shouldn't have missed that. This gave us a minor headache on Friday.

src/components/widget/SelectionDropdown.js
```diff
@@ -14,19 +15,7 @@ export default class SelectionDropdown extends Component {
         size: PropTypes.number.isRequired,
       }),
     ]).isRequired,
-    selected: (props, propName) => {
```
siemiatj: @PabloSichert can you throw some light on this part ? It looks like a huge overcomplication for something that should be pretty straightforward. As far as I know we expect `selected` to be an object, but I might be missing something. This on the other hand looks like it wants to handle different cases and work for different props (thus the propName I suppose) but will only work for `selected` prop and from my testing it always gets an object or `null`.

PabloSichert: It checks if `selected` is either one of `options` or `null/undefined`. So where you got the errors, something different was passed in, which should prevent the component from working properly.

No, it's only supposed to work for "selected" - you are right in that it should rather have been hardcoded there

PabloSichert:
> Removing onOutsideClick handler was an obvious thing and shouldn't have missed that.

I removed it since it wasn't needed anymore after refactoring. Could you give me a link on w101 where functionality is broken? I might have missed it

siemiatj:
@PabloSichert it was not possible to select anything from the dropdown or close them with mouse.

PabloSichert:
> it was not possible to select anything from the dropdown or close them with mouse.

I can't reproduce that. Could you give me a link where mouse interaction is broken?

siemiatj:
https://w101.metasfresh.com:8443/window/540345

```
Pick -> Pick to new HU -> Click on Packing item -> click outside => dropdown keeps opening
```

Issue #1641, opened 2018-03-01
Combined Lookup Widget Layout Glitch


metas-mk:
### Is this a bug or feature request?
Bug
### What is the current behavior?
Currently, the Lookup widget dropdowns seem sometimes to not be properly aligned with the corresponding field.
#### Which are the steps to reproduce?
- Open Sales Order window
- Record a new Sales Order for "Machine Parts"
- See Lookup dropdown in location/ User.
### What is the expected or desired behavior?
Shall be aligned under field and with same left border.

![image](https://user-images.githubusercontent.com/13365687/36845347-ab7c9300-1d56-11e8-9a1e-d74286f3daee.png)

PabloSichert:
With #1685 the dropdown will be as wide as the actual input field.

It would look like this:

![image](https://user-images.githubusercontent.com/4450694/37572060-e60e33fe-2b05-11e8-93fa-91f513ec041f.png)

Is that alignment okay?

metas-rc:
Results of IT


Tested on webui


* Currently, the allignment is like this:

![image](https://user-images.githubusercontent.com/16001442/37967958-894f4438-31d5-11e8-960e-37db5d0efa13.png)


Is this ok? 
@metas-mk  Please, move the task to done if you think it's ok.

metas-mk:
Quick check locally. Shall be fixed now. Pls. check in IT again.

metas-rc:
Results of IT

Tested on webui


* OK: Now the lookup widgets for sales order are perfectly alligned horizontally and vertically.

* Regression
* OK: In the orderline batchentry, the product and ifco are also alligned correctly
* OK: In the orderline advanced edit, all the fields that I checked are also alligned corrrectly.
* OK: The fields of sales order line look also alligned correctly if checked in grid view.

Issue #1586, opened 2018-02-12
Screen aliginment Improvements


metas-mk:
### Is this a bug or feature request?
Feature Request
### What is the current behavior?
The field label/content alignments are not very nice on screens at the moment.
#### Which are the steps to reproduce?
Open window/143 and see.
### What is the expected or desired behavior?
Align the field label/ content in a proper way.

Here a Screenshot with a few examples:
![image](https://user-images.githubusercontent.com/13365687/36098370-fece989c-0ffe-11e8-906a-19a129a34d2a.png)

metas-rc:
Results of IT


Tested on wbeui

* OK: The main view in sales order seems perfectly alligned. I dragged some right lines to check: 

![image](https://user-images.githubusercontent.com/16001442/38198349-100b503e-3695-11e8-8a7a-0f6d47461f63.png)




Regression: 


* Advanced edit also looks fine: 

![image](https://user-images.githubusercontent.com/16001442/38198458-9b921098-3695-11e8-88bd-b09a9f1045f1.png)

metas-rc:
Results of IT


Tested on webui


* OK: All looks nice now, including attributes and IsActive buttons

![image](https://user-images.githubusercontent.com/16001442/38369038-7d4f284e-38ef-11e8-8b72-660a46691690.png)

![image](https://user-images.githubusercontent.com/16001442/38369139-b92d9594-38ef-11e8-970b-1ed6bdd89977.png)

PR #1714, opened 2018-03-27
Cleanup input styles

siemiatj:
Tested on `window/143/xx`, `window/123/xx` and location modal. 

Related to #1586 & #1641


siemiatj:
screen shot 2018-03-28 at 12 45 40 pm

PR #1715, opened 2018-03-28
Fix lookup field input

siemiatj:
There was a bit of a mess with how and when shortcuts were disabled + shortcuts components were updating for no reason 10-20x on each click. 

Relates to #1555


src/components/keyshortcuts/ShortcutProvider.js
```diff
@@ -56,6 +56,8 @@ const codeToKey = {
   89: 'y',
   90: 'z',
   91: 'Meta',
+  109: '-',
+  107: '+',
```
PabloSichert: Just a nitpick, but those could go sorted numerically

src/utils/index.js
```diff
@@ -14,3 +14,13 @@ export const getQueryString = query =>
       return parameters;
     }, {})
   );
+
+export const arePropTypesIdentical = (nextProps, currentProps) => {
```
PabloSichert: Would rename `arePropTypesIdentical` -> `arePropsIdentical` since this is not about the types.

Alternatively, could name it more generic - `shallowCompare`

siemiatj: it is about the types. Check the function : 

```
if (typeof nextProps[key] !== typeof currentProps[key]) {
```

it has a limited use, but in some cases when a component is getting only functions in props we don't want to update it when nothing really changed.

siemiatj:
Ha, this is weird. I think git did some weird shit.

src/utils/index.js
PabloSichert: Oh, right. I completely overread that 😄

I think it would be a better idea to fix it by not passing in arrow functions from the parent. The current solution will have false positives when the function actually changed

PR #1719, opened 2018-03-29
Fix errors in profile section

siemiatj:
When debugging some profile-related bug I got errors in the console because Set object was not yet available when the app was initialized. Used `immutablejs`'s Set instead.

This is the page that was throwing the error: /window/53100/2188223. It should just show 404 instead.

screen shot 2018-03-29 at 7 57 44 pm

### DIFF
--- a/src/actions/WindowActions.js
+++ b/src/actions/WindowActions.js
@@ -5,6 +5,7 @@ import SockJs from 'sockjs-client';
 import currentDevice from 'current-device';
 import Stomp from 'stompjs/lib/stomp.min.js';
 import Moment from 'moment';
+import { Set } from 'immutable';
 
 import {
   ACTIVATE_TAB,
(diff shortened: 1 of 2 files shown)


src/actions/WindowActions.js
```diff
@@ -536,7 +537,7 @@ export function initWindow(windowType, docId, tabId, rowId = null, isAdvanced) {
               includedTabsInfo: {},
               scope: 'master',
               saveStatus: { saved: true },
-              standardActions: {},
```
PabloSichert: I think the problem was here - it was a plain object, so `{}.has` was `undefined` instead of being [`Set.prototype.has`](https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/Set/has). [`Set`](https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/Set) is natively supported from the global object

siemiatj: It was even weirder. Cause when I logged `Set` it was printed as undefined initially. Maybe it's some webpack bullshit.

src/reducers/windowHandler.js
```diff
@@ -297,15 +297,15 @@ export default function windowHandler(state = initialState, action) {
           },
         },
       });
-    case UPDATE_DATA_PROPERTY: {
```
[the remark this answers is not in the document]
siemiatj: Whoops, no it's not. No idea what I was thinking.

Issue #1606, opened 2018-02-16
Searching in dropdown fields


metas-lc:
### Is this a bug or feature request?
fr
### What is the current behavior?
we can't write anything in dropdown fields. Like e.g. want to search for something that starts with M , you have to scroll 
#### Which are the steps to reproduce?
1. open e.g. https://w101.metasfresh.com:8443/window/540173/1000002
2. select document base type
3. start writing M
=> you cannot write
### What is the expected or desired behavior?
if you write M, scrollbar (and selection) shall go to the word that starts with M "Maintenence order" in this case. If you write "Mat" => Match invoice

metasnw:
you also can not use the scroll icon as a workaround because once you click on it the dropwdown closes

![image](https://user-images.githubusercontent.com/16079655/36395737-ed52d1c0-15ba-11e8-82cc-42854d85a32a.png)

metas-lc:
note: we have a separate task for scrolling in dropdown: https://github.com/metasfresh/metasfresh-webui-frontend/issues/1605

metas-lc:
1. repeated the testcase above: pressed on M in document base type
=> i was sent to the first entry starting with M. Pressing M multiple times, would move to next ones starting with M
2. pressed other letters => the entries starting with those letters were selected OK
(for me it's fine how is working atm. Just to make sure: don't we want to be able to write stuff like "Mat"?)
3. if i click on the scrollbar from the dropdown, it closes NOK

siemiatj:
That's how it's implemented in the code. I think if we wanted to support longer text, we would have to show some input with the typed text.

metas-lc:
ok, if it will be needed i'll create another task for typing in dropdown fields

metas-lc:
IT
checked scrolling by clicking on scrollbar in dropdowns: header, filter, modal, gridview - the view is not closing anymore OK

PR #1724, opened 2018-04-03
Fixes for the lookup and list widgets

siemiatj:
This fixes : 
- search in the lookup field (no ticket)
- preventing closing the dropdown when clicking on the scrollbar #1606
- fixing error in the console #1723

### DIFF
--- a/src/components/widget/SelectionDropdown.js
+++ b/src/components/widget/SelectionDropdown.js
@@ -155,17 +155,17 @@ export default class SelectionDropdown extends Component {
           navigate(false);
           break;
         case 'Escape':
+          event.preventDefault();
           onCancel();
           break;
         case 'Enter':
+          event.preventDefault();
           onSelect(selected);
           break;
         default:
           return;
       }
     }
-
-    event.preventDefault();
   };
 
   handleKeyUp = () => {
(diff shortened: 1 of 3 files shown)


src/components/widget/SelectionDropdown.js
```diff
@@ -155,17 +155,17 @@ export default class SelectionDropdown extends Component {
...
         case 'Enter':
+          event.preventDefault();
           onSelect(selected);
           break;
         default:
           return;
       }
     }
-
-    event.preventDefault();
```
PabloSichert: Could you explain how this fix works? At least from what I see the only thing changed is the order of calling `preventDefault`.

Should `preventDefault` also not be added after [here](https://github.com/metasfresh/metasfresh-webui-frontend/blob/27b2967090f29f964eb3e03e0735e6b01a6cb9f3/src/components/widget/SelectionDropdown.js#L146)?

PabloSichert: I was asking because I can't see anything that would depend on the order of when `preventDefault` is called, as long as it happens synchronously in the handler. Am I wrong here?

> Nope. We want tab to work, and we want char key presses to be bubbled up to parent components.

There is a `return` in the `default` case, so all key presses that are not handled do not get a `preventDefault` and bubble up

siemiatj: Yup, and that's exactly what we want. Everything APART of those keys should be handled by parent components (if available) as we don't care about them here.

By the order I meant the how the events are handled by DOM.

PR #1729, opened 2018-04-04
Show selected emails in autosuggest dropdown

siemiatj:
There were also errors in the console caused by recent changes in the List components.

Relates to #1539

### DIFF
--- a/src/assets/css/email.css
+++ b/src/assets/css/email.css
@@ -196,7 +196,6 @@
 }
 
 .react-autosuggest__suggestions-list li {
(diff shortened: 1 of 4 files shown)


siemiatj:
I'd say both of those things are outside of the scope of the original ticket, and probably it would be better to have a separate ticket for this ?

metas-mk:
@siemiatj not really. The requirement was

> Show the selected items w/ similar look&feel to all other list items when selected (Lookup widget, List drop-down)

But I can create a follow-up issue for this one.

Issue #1752, opened 2018-04-19
Shortcuts not working on Mac


siemiatj:
### Is this a bug or feature request?
Bug
### What is the current behavior?
Keyboard shortcuts are not working. Probably due to the fact, that `Alt` on Mac is `Option/Alt`.
#### Which are the steps to reproduce?
Try any of the shortcuts when running on Mac.
### What is the expected or desired behavior?
Shortcuts should work.

siemiatj:
Yeah, but I think we should have a separate ticket for this. On OSX shortcuts don't work at all.

metas-dh:
Results of IT1
tested in OS X 10.11.6

shortcuts working acc. to http://docs.metasfresh.org/webui_collection/EN/Keyboard_shortcuts_reference.html:

* not working are ```⌥ alt + W``` for Clone and ```⌥ alt + R``` for Letter, but those have not been working at all for a while, see #1410
* the rest work fine: OK

metas-ts:
:cherries: picked to `release`

PR #1753, opened 2018-04-19
Fix shortcuts on OSX

siemiatj:
Tested on Windows machine as well. Seems to be working fine.

Related to #1752


siemiatj:
@metas-mkI just checked this branch with w101 and it seems to be working exactly the same (apart of the typing bug). [alt] + + expands the table, and [alt] + u throws some error.

Fixed the bug with typing by removing the modifier keys from the sequence. Good catch there.

PR #1754, opened 2018-04-20
Plugins support

siemiatj:
Work in progress.


siemiatj:
Looks like this is ready to test. @metas-ts or @teosarca can you guys build a prod instance using this branch and check if everything works fine (build-wise ) ? I've updated a lot of dependencies and updated the build configs + configs itself, so I want to be sure this won't break anything :)

metas-mk:
@siemiatj triggered a dev instance https://dev44.metasfresh.com/ but, does not show the login screen.

siemiatj:
Fixed some small issues and console warnings, now looks to be stable. If the dev instance looks fine I think we can merge this on Monday.

PR #1760, opened 2018-04-24
Fix lookup fields

siemiatj:
When clicking the clear button on a nested lookup field, default values are still displayed even though only the first field is editable.


siemiatj:
@metas-mk ready to re-test. Lookups really need refactoring, haven't had time to do this yet.

Issue #1410, opened 2017-12-05
Shortcuts for Clone & Letter not working


metas-dh:
### Is this a bug or feature request?
Bug
### What is the current behavior?
Shortcuts for Clone & Letter are not working. Home button for jumping to 1st page also doesn't work.
#### Which are the steps to reproduce?
Try [alt]+w ( [alt]+r, in e.g. window 143:

![2017-12-05 11_49_29-metasfresh](https://user-images.githubusercontent.com/18696715/33605902-daabdbc4-d9bb-11e7-8dd2-3f448843fd2a.png)

### What is the expected or desired behavior?
Shortcuts shall work.

metas-dh:
Update:
```Pos1``` (Home button) working fine, see #1663

metas-lc:
IT
checked Try [alt]+w and [alt]+r => they work fine  OK
(checked on new and existing documents, in sales order)
also checked if home button is still working OK

PR #1766, opened 2018-04-27
Fix CLONE and LETTER global shortcuts

siemiatj:
Will have to create a follow up tickets, because cloning throws a `method not supported` error, and letter modal needs similar fixes to email modal.

Related to #1410 #1768


siemiatj:
Ready to retest.

metas-mk:
Checked locally. Works as required. Thanks. Merging.

PR #1793, opened 2018-05-16
Refresh view after removing a static filter

siemiatj:
Related to #1790

### DIFF
--- a/src/api/view.js
+++ b/src/api/view.js
@@ -91,13 +91,9 @@ export function filterViewRequest(windowId, viewId, filters) {
 
 export function deleteStaticFilter(windowId, viewId, filterId) {
   return del(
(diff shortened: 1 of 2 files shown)


teosarca:
works ok but not when u are already in same window.
i.e.
image


* open http://localhost:3000/window/540375
* go to notification and click on that notification which will open a view with same window ID
=>
the URL is changed (OK), but nothing else.

siemiatj:
I think this comment should be for https://github.com/metasfresh/metasfresh-webui-frontend/pull/1792. Anyways will fix in a separate PR.

teosarca:
right :) i will integrate this PR and will wait for the other PR.
also i will copy the comment to #1789 for our testing team.

10x

PR #1803, opened 2018-05-22
Fix editing modal view

siemiatj:
Bunch of things was wrong there - fields constantly in edit mode + stuff from the ticket. This is the first time we're actually altering the UI on response from the server instead of just re-rendering everything.

It looks good to me but if someone could check if editing table cells works as expected - that'd be great.

Issue #1801, opened 2018-05-22
Error when opening HU Editor window / selecting a HU in picking tray clearing


metas-dh:
### Is this a bug or feature request?
Bug
### What is the current behavior?
When opening HU Editor window (window/540189), i got

```
Error: For input string: "[object...
Server error
For input string: "[object Object]"

{timestamp: 1526993588371, status: 500, error: "Internal Server Error",…}
error
:
"Internal Server Error"
exception
:
"java.lang.NumberFormatException"
message
:
"For input string: "[object Object]""
path
:
"/rest/api/documentView/packingHUs/packingHUs-CK-2156425-2205175/quickActions"
status
:
500
timestamp
:
1526993588371
trace
:
["java.lang.NumberFormatException: For input string: "[object Object]"",…]
[0 … 99]
0
:
"java.lang.NumberFormatException: For input string: "[object Object]""
1
:
"	at java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)"
2
:
"	at java.lang.Integer.parseInt(Integer.java:580)"
3
:
"	at java.lang.Integer.parseInt(Integer.java:615)"
4
:
"	at de.metas.ui.web.handlingunits.HUEditorRowId.fromJson(HUEditorRowId.java:110)"
5
:
"	at de.metas.ui.web.handlingunits.HUEditorRowId.ofDocumentId(HUEditorRowId.java:77)"
6
:
"	at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:193)"
7
:
"	at java.util.Iterator.forEachRemaining(Iterator.java:116)"
8
:
"	at java.util.Spliterators$IteratorSpliterator.forEachRemaining(Spliterators.java:1801)"
9
:
"	at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:481)"
10
:
"	at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:471)"
11
:
"	at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:708)"
12
:
"	at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)"
13
:
"	at java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:499)"
14
:
"	at de.metas.ui.web.window.datatypes.DocumentIdsSelection.toSet(DocumentIdsSelection.java:277)"
15
:
[...]

PR #1804, opened 2018-05-23
Fix HU editor

siemiatj:
We really need to make it easier on the frontend to differentiate between the views or send API url's with the widget data.

**Notice**

This requires #1803 to be merged first.

Related to #1800 #1801

### DIFF
--- a/src/components/widget/List/List.js
+++ b/src/components/widget/List/List.js
@@ -103,7 +103,7 @@ class ListWidget extends Component {
 
     let request = null;
 
-    if (viewId && !filterWidget) {
+    if (viewId && entity === 'window' && !filterWidget) {
       request = dropdownModalRequest({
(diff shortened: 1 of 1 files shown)


teosarca:
@siemiatj  @metas-mk  quick checked. Works nice on HU editor, but on regular views, the quick actions are not correctly queried for following tst case:
* open Material Receipt Candidates: http://localhost:3000/window/540196
* remark: no rows are selected => OK, the `/quickActions` without param is called => OK (because we have no selection)
* select first row
* remark: the `/quickActions?selectedIds=<that row>` endpoint is called => OK
* now click on breadcrumb `Material Receipt Candidates` link
* remark:
** view is refreshed => OK
** first row remains selected => OK
** `/quickActions` (without selectedIds param) is called => NOK
image

Issue #1814, opened 2018-05-29
Action button is too large


metas-lc:
### Is this a bug or feature request?
bug
might be related with https://github.com/metasfresh/metasfresh-webui-frontend/pull/1754
### What is the current behavior?
action button is too large
#### Which are the steps to reproduce?
1. open bpartner pharma window and see (id 540409)
![image](https://user-images.githubusercontent.com/16001666/40660506-d76b2014-6359-11e8-8f60-8c4af8d885b1.png)

### What is the expected or desired behavior?
buttons should not expand if the title is too long

metas-mk:
For IT: Please also check the Batch Entry Layout in Sales Order/ Purchase Order if ok.

metas-mk:
for IT: PR https://github.com/metasfresh/metasfresh-webui-frontend/pull/1818 is now merged

metas-lc:
IT
1. button doesn't expand anymore (you can hover to see the whole name) OK
2. Batch Entry Layout in Sales Order/ Purchase Order - OK

PR #1818, opened 2018-05-30
Fix actions buttons size

siemiatj:
Related to #1814


metas-mk:
Merged lastest master because of PR revert
image

PR #1823, opened 2018-06-01
Reload translations after login

siemiatj:
If language stored on the user modal has changed we will fetch for updated messages.
This PR also fixes logout error.

Related to #1616


metas-ts:
:cherries: picked to release

Issue #1808, opened 2018-05-24
Shortcuts for inbox, menu and sidelist


metas-lc:
### Is this a bug or feature request?
bug
### What is the current behavior?
alt+3, alt+4, alt+5, alt+6 and alt+7 when are pressed twice, should close the lists. Now is just refreshing them
#### Which are the steps to reproduce?
e.g. open order window, press alt+3 (inbox), alt+4 (menu), alt+5 (doc list), alt+6 (ref docus) or alt+7 (attachments) twice
=> the menus are not closing, but refreshing

### What is the expected or desired behavior?
when pressed twice, shall close

metas-lc:
please also check how the following shall work: 
when you are in e.g. letter (alt+r) and press alt+d, delete popup appears in the background OK?

PabloSichert:
> when you are in e.g. letter (alt+r) and press alt+d, delete popup appears in the background OK?

I'd say it's fine. Otherwise one could confuse the dialog with deleting the letter only, when it actually deletes the whole order.

metas-lc:
IT
alt+3, alt+4, alt+5, alt+6 and alt+7 when are pressed twice, close the lists OK

PR #1825, opened 2018-06-02
Toggle header items when hitting the same hotkey repeatedly

PabloSichert:
#1808

### DIFF
--- a/src/components/header/Header.js
+++ b/src/components/header/Header.js
@@ -111,6 +111,10 @@ class Header extends Component {
     this.setState({ isUDOpen: !!state });
   };
 
+  handleUDToggle = () => {
+    this.setState({ isUDOpen: !this.state.isUDOpen });
+  };
+
   handleMenuOverlay = (e, nodeId) => {
     const { isSubheaderShow, isSideListShow } = this.state;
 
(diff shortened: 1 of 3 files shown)


src/components/header/Header.js
```diff
@@ -280,32 +288,39 @@ class Header extends Component {
...
+      isUDOpen: false,
+      isSideListShow: false,
+      tooltipOpen: '',
+    };
+
+    if (clickedItem) {
+      delete state[clickedItem];
+    }
+
+    state.isSubheaderShow =
```
siemiatj: isn't omitting `setState` considered a bad practice in the react world ?

src/components/keyshortcuts/GlobalContextShortcuts.js
```diff
@@ -1,16 +1,51 @@
 import React, { Component } from 'react';
+import PropTypes from 'prop-types';
```
siemiatj: 👍

src/components/header/Header.js
```diff
@@ -103,10 +103,18 @@ class Header extends Component {
...
 
+  handleInboxToggle = () => {
+    this.setState({ isInboxOpen: !this.state.isInboxOpen });
+  };
+
   handleUDOpen = state => {
     this.setState({ isUDOpen: !!state });
   };
 
+  handleUDToggle = () => {
```
siemiatj: Can you add a comment what UD stands for ? I assume it's User's Dropdown ?

src/components/header/Header.js
siemiatj: my bad, carry on ;)

src/components/header/Header.js
PabloSichert: I have no idea, just went with the existing naming in the code. That illustrates your point well, though 😄

Issue #1828, opened 2018-06-04
Truncate action names with ellipsis


metas-lc:
### Is this a bug or feature request?
bug
### What is the current behavior?
Action names are aligned to left and are not truncated when the names are too long
#### Which are the steps to reproduce?
1. login with de_DE, open e.g. receipt candidates window 540196, select a row with HU
![image](https://user-images.githubusercontent.com/16001666/40916150-524d42a6-6807-11e8-80fd-b4c622e60922.png)

=> notice that the name doesn't fit
=> it's a bit visible on next line
=> not truncated
=> not centrally aligned

### What is the expected or desired behavior?
e.g. of how it used to be: 
![image](https://user-images.githubusercontent.com/16001666/40916194-82e5ed32-6807-11e8-82fe-4ee5a465edb8.png)

align text to center, truncate long names with ellipsis

metas-dh:
Results of IT1
tested in mf15

* long action names are truncated: OK
* texts are aligned to center: OK
* action button has the correct size (see #1814): OK

metas-ts:
:cherries: picked to `release`

PR #1829, opened 2018-06-04
Fix quick action names alignment/truncating

siemiatj:
Related to #1828


metas-ts:
:cherries: picked to `release`

Issue #1686, opened 2018-03-19
API login for cypress


siemiatj:
### Is this a bug or feature request?

Feature

### What is the current behavior?

Right now it's not possible to authenticate a user without going through the whole login form flow, as we're setting some variables in the internal state on form response.

#### Which are the steps to reproduce?

Try using `loginByForm` custom command.

### What is the expected or desired behavior?

We should be able to login without using the login form.

siemiatj:
@metas-lc I think we can close this one since there's nothing to test here.

metas-lc:
@siemiatj Thanks. Closing it

PR #1835, opened 2018-06-08
Implement API login command for cypress tests

siemiatj:
This also fixes errors in the console caused by incorrect (or actually missing) propTypes arguments.

Command itself is not perfect, as it shows cypress errors for aborted requests (when redirecting) but it's enough for us to get started on adding some tests coverage.

Related to #1686


siemiatj:
@metas-ts done !

Issue #1467, opened 2017-12-19
Advanced Edit shortcut not working correctly


metas-dh:
### Is this a bug or feature request?
Bug
### What is the current behavior?
Advanced Edit shortcut does not work after you changed sth / clicked somewhere in the resp. window.
#### Which are the steps to reproduce?
1. Open a window, use [alt]+e directly: Advanced Edit view opens, OK
2. Open another window, change sth or just click somewhere in the window
3. Use [alt]+e again: Advanced Edit view does not open, you have to use Actions to get there: NOK
### What is the expected or desired behavior?
Shortfut [alt]+e shall always open Advanced Edit view.

metas-lc:
IT
1. open a window, press alt+E => advanced edit opens OK
2 .open a document, select a line, press alt+e => advanced edit opens for the line OK
3. open another window, do some changes /clicks, press alt+e => advanced edit opens OK
4. do some changes in a line, select that line, press alt+e => advanced edit opens OK

Issue #1527, opened 2018-01-23
Shortcut [alt]+l toggling the wrong quick actions menu when 2 modal windows


metas-dh:
### Is this a bug or feature request?
Bug
### What is the current behavior?
When there are 2 modal windows, e.g. in picking tray clearing, shortcut [alt]+l  toggles the wrong quick actions menu.
#### Which are the steps to reproduce?
1. Have HUs picked and processed
2. Open Picking Tray Clearing 
3. Select the HU you picked and processed, in the modal window on the left
4. Use shortcut [alt]+l : => Quick actions menu for the modal window on the right is displayed: NOK
### What is the expected or desired behavior?
Quick actions for the window where sth is selected shall be displayed.

siemiatj:
![nfp2tf96l5](https://user-images.githubusercontent.com/513460/42214801-cfd8a794-7ebc-11e8-85ec-69f393e2f418.gif)

I see two problems here : 
- (after #1847 is merged) we push the shortcut to the to the latest element at the beginning of the queue, as in 99.99% of the time that's the element we're interested in (think modals). So we'd have to somehow control this in the code
- even if ^^ would not be true, there's a selection in the right column as well

PR #1847, opened 2018-06-13
Fix fire shortcut handlers when defined in multiple places

siemiatj:
This requires some proper testing, as it potentially might cause unexpected behaviour. Right now we were always trying to fire only the first shortcut handlers, even if we had more than one defined (like the case of `Advanced Edit` mentioned in the issue). Instead we should try calling all of them and handle which one is available at which point in the code. In case mentioned in #1467, there's a global `Advanced Edit` and a local for the Table component.

**UPDATE:**
I've found a way to fix this without introducing hacks, so should be fairly safe. Unless we have other duplicated shortcuts (in which case we'd have to add a return values in the code to control which one is fired and when) this should just work.


siemiatj:
Fixing CI errors

siemiatj:
Worth testing #1527 as it might be related.

PR #1870, opened 2018-07-04
Allow rendering frequent filters inline

siemiatj:
Related to #1859 

TODO:
[ ] there's some backend error when blurring the field


siemiatj:
screen shot 2018-07-10 at 18 31 35

PR #1896, opened 2018-07-26
Use translated texts and fix sales order tests

siemiatj:
Now instead of hardcoded texts (which would break when other languages were used) we can use translations and captions from data responses.

### DIFF
--- a/cypress/integration/sales_order_spec.js
+++ b/cypress/integration/sales_order_spec.js
@@ -1,21 +1,50 @@
+import config from '../config';
+
 describe('New sales order test', function() {
+  const windowId = 143;
+  let caption = '';
+  let nodeId = null;
+  let menuOption = '';
+
   before(function() {
-    // login before each test
     cy.loginByForm();
-    cy.visit('/window/143');
+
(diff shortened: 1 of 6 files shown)


cypress/integration/sales_order_spec.js
```diff
@@ -1,21 +1,50 @@
...
+        caption = response.body.captionBreadcrumb;
+        nodeId = response.body.nodeId;
+
+        cy.request('GET', config.API_URL+'/menu/node/'+nodeId+'/breadcrumbMenu')
+          .then((response) => {
+            const resp = response.body;
+            expect(resp.length).to.be.gt(0);
+
+            for (let i=0; i<resp.length; i+= 1) {
+              if (resp[i].nodeId === '1000040-new') {
```
metas-ts: what is the `1000040` about? is it a constant value? i'm not sure, it might be incremenented on each new order..

siemiatj: I think it's not but I just changed it to look for `-new`.

metas-dh:
checked the tests, for test system with menu de_DE and en_US, worked fine for both: OK

PR #1897, opened 2018-07-27
Support multiple lookups in one field group

siemiatj:
Related to #1889 . ~Seems to be working fine but I want to double check it tomorrow.~


teosarca:
quick checked: no PATCH is sent for the third field:
image

Issue #1901, opened 2018-07-30
Highlighted action is not the one from mouse over when calling actions with [alt]+l


metas-dh:
### Is this a bug or feature request?
Bug
### What is the current behavior?
http://recordit.co/hOWQDSBq3H

This happens when there are 2 sides of an HU Editor overlay, e.g. in manufacturing window/53009 (see video) and also in picking clearing terminal, window/540371. Does not happen when actions are called with mouse. Clicking the action from mouse over works fine too.
#### Which are the steps to reproduce?
See video.

siemiatj:
Is/was it ever calling highlighted action and not some predefined default ? I was sure it's the latter, especially that highlighting something with the mouse and then using keyboard to execute it seems unintuitive.

siemiatj:
Ok if I understand correctly, PR should fix it.

http://recordit.co/Ze8TIYaURS

metas-lc:
IT
1. Picking Tray Clearing (Prototype). select a line on right side, press alt+L, hover the actions with mouse
=> the action you are with the mouse on that moment gets grey OK
2. Manufacturing order. Select co-product, HU Aus.. button. Press alt+L, hover the actions with mouse
=> the action you are with the mouse on that moment gets grey OK

(also used mouse click instead of alt+L. Works the same OK)

PR #1906, opened 2018-08-01
Handle quickactions shortcuts with included views

siemiatj:
When there's included view and a valid selection, quickactions will first react to the included view's shortcut.

Related to #1901


metas-ts:
hi @siemiatj codacy is whining; please check if it has a point

Issue #1909, opened 2018-08-01
Vertical Cabling : Show full product name on quick order entry 


metasnw:
### Is this a bug or feature request?
FR

### What is the current behavior?
its cut off
![image](https://user-images.githubusercontent.com/16079655/43460146-1939ef0a-94d0-11e8-88e6-f8ebe5bfe0a8.png)

#### Which are the steps to reproduce?
1. add a product in order line with a name > 10 digits

### What is the expected or desired behavior?
name should be displayed in full length for all 3 fields

PR #1910, opened 2018-08-02
Changes to quickactions batch entry

siemiatj:
Related to #1899 and #1909


siemiatj:
This codacy is so lame. Focusing on bullshit instead of real problems :)

Issue #1785, opened 2018-05-10
Search field in Pricing conditions - keyboard selection


metas-lc:
### Is this a bug or feature request?
bug
follow-up of https://github.com/metasfresh/metasfresh-webui-frontend/issues/1765
### What is the current behavior?
cannot select an entry using the keyboard
#### Which are the steps to reproduce?
1. go to ales order, create a new one, add a product. Select the line, go to actions and start Pricing Conditions
2. write smth in payment term  (e.g. press SPACE)
=> list with entries 
3. try to use down arrow to select an entry
=> the second row is selected instead NOK
### What is the expected or desired behavior?
the entries in the list should be selected

siemiatj:
Can't really reproduce this. It may be because I turned off the default edit state for all editable fields.  I opened the pricing conditions, selected the field, pressed enter and then changed options with keyboard arrows.

![errooo](https://user-images.githubusercontent.com/513460/40480897-1881f3c4-5f50-11e8-8841-579b59b41f0c.gif)

metas-lc:
Please try the same testcase, but instead of Type, use Payment Term (seems it;s happening on search fields)
if you have something in Payment Term press x and then press space and down arrow
=> it moves to next fields
![image](https://user-images.githubusercontent.com/16001666/41035425-47801030-6995-11e8-82f5-75f46b66f30d.png)

metas-lc:
IT
When i double click on Payment Term, the first field is selected. I can select the column and start writing, but if i select the scrollbar to get back to it, i have to double click again and i get back to the first field
http://recordit.co/zXJIKJvo8i
happens the same with other fields, but not exactly the same way: you e.g. double click on payment discount, and right after select the next field => jumps
.
checked search fields in other places: Product window, tab CU_TU allocations. If you start searching in a search field, you cannot close it by moving to other fields
![image](https://user-images.githubusercontent.com/16001666/42804184-e027ae5c-89b0-11e8-9534-54eb76fe78ed.png)
.
jumping also happens in other places: included tabs: see business partner window, change things in Customer tab. 
Same: product price window (540325), you can edit prices there with right click. If you select a field near, jumps.
.
in headers (e.g. product price window) search fields seem fine

siemiatj:
It is related. I think it's because I wanted to have focus on the table after closing the dropdown. But apparently I was too aggressive with this ;)

metas-dh:
[...]

PR #1915, opened 2018-08-07
Fixes for lookups

siemiatj:
Fix closing lookup list without select when clicked outside and scrolling through options with keyboard. Also commented out some code that seems to be irrelevant now. Leaving it for the time being in case it turns out this thing is used in some edge case.

Related to #1785


src/components/widget/Lookup/Lookup.js
```diff
@@ -22,8 +24,11 @@ class Lookup extends Component {
           fireClickOutside: false,
           fireDropdownList: false,
           isFocused: false,
+          isInputEmpyt: false,
```
metas-ts: minor typo

PR #1916, opened 2018-08-08
Fix barcode scanner in HU editor

siemiatj:
Last filter's changes broke this.


metas-dh:
Results of IT1
tested in mf15

Barcode filter working fine again (tested in manufacturing order and picking terminal):
* Barcode entry overlay opens correctly: OK
* Barcode filter works fine: OK

Issue #1932, opened 2018-08-26
Mandatory Field not filled Error too early


metas-mk:
### Is this a bug or feature request?
Bug
### What is the current behavior?
In screens with multiple mandatory fields, an error is shown to the user that mandatory are not filled. This happens after each patch.
#### Which are the steps to reproduce?
Create a new record in a window with multiple mandatory fields.
### What is the expected or desired behavior?
The error message shall not be shown as long as there are mandatory fields not filled that haven't been focused or changed initially.

siemiatj:
So I've noticed, that this seems to be triggered by lookup/list fields. I've started with a sales order and removed everything : 

![initial](https://user-images.githubusercontent.com/513460/44666750-5c0c1480-aa19-11e8-9021-6d6d254922bd.png)

Then I've started adding values to fields:

http://recordit.co/1uF9SoXly8

So as far as I understand:
- if business partner is empty and we're setting the target document type, the message about business partner should not be displayed (and a similar case - when warehouse and document type are both empty and we'll set the value for warehouse, we should not get the message for document type)
- red indicator line should stay until we successfully save tho document

correct ?

metas-dh:
Results of IT1
tested in mf15

1. error message shall not be shown as long as there are mandatory fields not filled that haven't been focused or changed initially: 
* changes are saved (green line in header) after setting each mandatory field: OK
* no error msg while filling the mandatory fields: OK

2. error message shown when trying to complete document before all mandatory fields are filled: 
* error msg shown with info about mandatory field not being filled: OK
* red line shown in header until missing (next) mandatory field was filled: OK

PR #1938, opened 2018-08-27
Hide unnecessary empty fields errors

siemiatj:
Related to #1932 .

Please let me know if this is how it's supposed to work.


metas-ts:
@metas-mk I tried it out locally. works

Issue #1948, opened 2018-08-31
Filter Attribute Label not set properly


metas-mk:
### Is this a bug or feature request?
Bugfix
### What is the current behavior?
Currently, the Filter Label in Button is not set properly.
#### Which are the steps to reproduce?
Open, try and see. /window/123
### What is the expected or desired behavior?
Shall be set properly.

![image](https://user-images.githubusercontent.com/13365687/44904246-ff597400-ad0e-11e8-91fc-19a49378c549.png)

metas-ts:
:cherries: picked commit 45fca75825781a4a4e333655320f85a6ba03a940 to `release`

metas-dh:
Results from e2e
tested in release

1. filter for name: "test" in bpartner window: OK
![2018-09-03 17_15_43-notificationsform](https://user-images.githubusercontent.com/18696715/44994058-2c22bb00-af9d-11e8-8f83-38153c0fd611.png)

2. filter for order bpartner "test" in sales order window: NOK i think
![2018-09-03 17_16_21-notificationsform](https://user-images.githubusercontent.com/18696715/44994062-32b13280-af9d-11e8-984d-1bd9a86ae69f.png)

metas-dh:
Results of IT1
tested in mf15

Filter Attribute Label set properly:

1. bpartner window: set Filter e.g. 
* Manufacturer: Y: filter label: Manufacturer: **OK**
* add Customer: Y: filter label: Kunde, Manufacturer: **OK**
* add Name: test: filter label: test, Kunde, Manufacturer: **OK**
* filter results matching: **OK**

2. shipment window: set Filter e.g.
* shipment to: Test Kunde, document status: Completed, org: metasfresh AG: filter label: Test Kunde 1_G0001, Fertiggestellt, ...: OK; mouse over filter label: all params are displayed: **OK**
* filter results matching: **OK**

3. Default filter, e.g. "Aktiv" set in bpartner B2C window (main window): 
* filter label shows "Aktiv": **OK**
* filter results matching: **OK**

4. Default filter, e.g. attribute filter, in modal overlay (=> picking terminal, HU selection window):
* filter label shows "Attribute berücksichtigen", but not like it is set (although it is set: **NOK**

![2018-09-12 10_57_33-notificationsform](https://user-images.githubusercontent.com/18696715/45414162-10917180-b67b-11e8-9d93-9299c3402f81.png)


[...]

PR #1963, opened 2018-09-12
Fixes for filters

siemiatj:
Reated to #1944 & #1948 

Fixes:
- propTypes error
- showing filter `Filter` in not-frequent filters panel even if nothing was set
- fix error in the console when resetting frequent filters
- fix error of showing inactive filter with defaultValue in caption which also blocked the filters list (issue on client's setup)

### DIFF
--- a/src/components/filters/Filters.js
+++ b/src/components/filters/Filters.js
@@ -50,7 +50,7 @@ class Filters extends Component {
             });
           }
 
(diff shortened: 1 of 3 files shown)


siemiatj:
I think we can merge this one to get rid of some small issues with frequent filters.

Issue #1952, opened 2018-09-06
Advanced tooltip support


metas-ts:
### Is this a bug or feature request?
FR

#### Widget with additional included fields

 this is the JSON layout info that the API provides today:

<details>

```json
	"sections": [{
		"columns": [{
			"elementGroups": [{
				"type": "primary",
				"elementsLine": [{
					"elements": [{
						"caption": "C_BPartner_ID | C_BPartner_Location_ID | AD_User_ID",
						"description": "Bezeichnet einen Geschäftspartner",
						"widgetType": "Lookup",
						"type": "primary",
						"size": "L",
						"viewEditorRenderMode": "never",
						"fields": [{
							"field": "C_BPartner_ID",
							"source": "lookup",
							"emptyText": "leer",
							"newRecordWindowId": "540327",
							"newRecordCaption": "Neuer Geschäftspartner",
							"supportZoomInto": true
						},
						{
							"field": "C_BPartner_Location_ID",
							"source": "list",
							"emptyText": "leer",
							"supportZoomInto": true
						},
						{
							"field": "AD_User_ID",
							"source": "list",
							"emptyText": "leer",
							"supportZoomInto": true
						}]
					}]
				},
				{
					"elements": [{
						"caption": "Bill_BPartner_ID | Bill_Location_ID | Bill_User_ID",
						"description": "Geschäftspartners für die Rechnungsstellung",
[..]
   }]
```

</details>

in the frontend, this look like this (more or less, note that in the screeshot, AD_User_ID is null): 
![order](https://user-images.githubusercontent.com/1264789/45138835-29e37b00-b1ae-11e8-8c8e-dcd7e283d772.png)

if e.g. the field `AD_User_ID` shall not be a `widget`, but `tooltip`.
how should the json look in that case..
idea: introduse a property (or is it called variable?) called type which can be widget or overlay.
in case of widget, everything stays the same.
in case of overlay ...

there is an additional property `icon` ... icon-values can be `text`, `help` and more...they refer to small images..e.g. there is a `tooltip-help.png` file somewhere in the webui-frontend.
so the API might return some JSON like:

<details>

```json
    "sections": [{
        "columns": [{
            "elementGroups": [{
                "type": "primary",
                "elementsLine": [{
                    "elements": [{
                        "caption": "C_BPartner_ID | C_BPartner_Location_ID | AD_User_ID",
                        "description": "Bezeichnet einen Geschäftspartner",
[...]

PR #1966, opened 2018-09-17
Tooltip for widgets

siemiatj:
A popup field that can be displayed in other widget fields. Triggered by clicking an icon. Currently supported in Lookups. Hides when clicked outside.

Related to #1952 

@metas-ts let me know if you want this to look/behave differently or add something more to it.

screen shot 2018-09-17 at 15 00 39

screen shot 2018-09-17 at 15 00 32


siemiatj:
Now widget will only be shown if there's a value defined.

PR #1967, opened 2018-09-17
Show description for field labels

siemiatj:
If description is not available, use caption as we do now. Related to #1953


siemiatj:
@metas-ts 🆗

PR #1971, opened 2018-09-18
Make it easier to work with tabs in cypress

siemiatj:
Related to #1926

### DIFF
--- a/cypress/.gitignore
+++ b/cypress/.gitignore
@@ -1,3 +1,3 @@
 screenshots
-
 videos
+/config.js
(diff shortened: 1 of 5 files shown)


siemiatj:
@metas-ts it's a really generic thing so I don't think creating a custom command makes sense. It should just work like this:

`cy.get('#internalName').click()`. 

But I just realized I added the id on the tab for selecting, and not the pill for clicking. Let me fix that.

siemiatj:
screen shot 2018-09-19 at 09 48 05

metas-ts:
>  It should just work like this:
> 
> cy.get('#internalName').click().

That's particularly great @siemiatj , buuut, afaiu that makes it really important to have the `tab_` prefix
I.e. afaiu we need to have 
`cy.get('tab_#internalName').click();`

to avoid colision with unrelated elements from a different part of the application dictionary that just happen to have the same `#internalName`

wdyt?

siemiatj:
Yeah, makes sense. And then creating a custom command makes sense as well :)

Issue #1989, opened 2018-10-03
Hourglass Overlay for pending Posts


metas-mk:
### Is this a bug or feature request?
Feature Request
### What is the current behavior?
Currently. when filtering in windows with a large amount of date, it sometimes takes a while for the result. The post is pending. After a short while the user believes that the filtering "apply" did not work and tries again.
#### Which are the steps to reproduce?
Open, try and see.
### What is the expected or desired behavior?
Add an Hourglass overlay feature for posts that are pending longer than for example 3secs. This shall show the user that something is happening,

metas-dh:
Results of IT1
tested in sp80 e2e

* Spinner pops up when filtering takes a while (>2 sec.): OK
* Spinner also pops up diretly when filtering in windows with **lots** of data (OLCands): afais that's OK

metas-ts:
:cherries: picked  https://github.com/metasfresh/metasfresh-webui-frontend/issues/1989 into branch `5.76_hf1`

PR #2010, opened 2018-10-16
Spinner overlay

siemiatj:
For longer operations we should show a spinner. Idea was to initialize it after a configurable delay, but in some cases this won't work as the UI is blocked by heavy operations and React won't update anything (picking terminal is a good example where preparing over 2800 nested table rows just takes a really long time). So by default I set the delay to 300ms which allows it to render everywhere. Also usually spinners are shown immediately so I think this should work for us.

Related to #1989 

screen shot 2018-10-17 at 01 49 15


siemiatj:
Test crashing, fixing.

PR #2026, opened 2018-10-25
Run process action after scanning QR code

siemiatj:
Looks like they were never run for any overlay fields but I might be wrong here. Related to #2015


teosarca:
quick tested: got "fill mandatory param BArcode"... that's happening because the param is PATCHed after the /start (process) is called:

image


Also, pls note that the PATCH is called twice.

siemiatj:
Well, actually they're patched before the request to `/start` but because Javascript, it ends up being a huge mess. Looks like MasterWidget component needs some love.

teosarca:
quick tested again, but got the same issue

image

siemiatj:
As far as I understand how this works this is correct, cause this barcode just doesn't exist. Correct me if I'm wrong.

teosarca:
@siemiatj  yes, but take a look at that picture.
The second(OPTIONS) and third(PATCH) calls shall be executed before first call (/start).
Because we need to communicate the up2date params before starting the process.

So it's not about the error message, you will get the error anyways because the param is not correct.
It's about the fact that you start the process with old params and after that frontend is communicating the params.

teosarca:
@siemiatj  please make sure the action can be executed when using a tablet

siemiatj:
@teosarca can you check if the order is correct now ?

PR #2051, opened 2018-11-05
Fetch quickActions when selection range changes

siemiatj:
Only reproducible on mobile/with ranges. Related to  #2049.


siemiatj:
@teosarca should be good now

teosarca:
now tablet toggle selection got broken: https://github.com/metasfresh/metasfresh-webui-frontend/issues/2006

i.e. when you click o a row, the previous row got deselected.

image

siemiatj:
Ok so I tested and it works both on mobile and desktop acting as mobile. The problem is, for some dumb reason I agreed to remove pagination on mobile devices and now the browser just can't handle the amount of rows we're trying to show.

PR #2052, opened 2018-11-05
Fetch actions after updating rows

siemiatj:
Related to #2050


teosarca:
got following error when i tried to add a new sales order line using batch entry

image

teosarca:
happens on master too: https://github.com/metasfresh/metasfresh-webui-frontend/issues/2054

teosarca:
quick tested, works nice

PR #2084, opened 2018-11-27
cypress pressDoneButton fails if the respective record was not saved

metas-ts:
Note: without this, cypress will by default confirm, so the modal will close without an error and the record won't be inserted or updated

See `window:confirm` at https://docs.cypress.io/api/events/catalog-of-events.html#App-Events

https://github.com/metasfresh/metasfresh-webui-frontend/issues/2083


siemiatj:
👍

Issue #2082, opened 2018-11-27
Show tab description as tool tip


metas-ts:
### Is this a bug or feature request?
FR
Analog to https://github.com/metasfresh/metasfresh-webui-frontend/issues/1953

### What is the current behavior?

![image](https://user-images.githubusercontent.com/1264789/49065102-4d521b00-f21d-11e8-9f9b-736455d901c2.png)

#### Which are the steps to reproduce?

* Hover over a tab whose `AD_Tab` does have a description in our application dictionary

### What is the expected or desired behavior?

The description should be shown

metas-dh:
Results of IT1
tested in mf15


* if there is a  tab description, this is shown as tool tip: OK
* if there is no tab description, the tab's caption is shown: OK
* description / caption shown acc. to menu language: OK

PR #2088, opened 2018-11-28
Add tabs description in tooltip

siemiatj:
Related to #2082


siemiatj:
No idea how this worked for me before without those changes. Probably some leftovers from other branch. Fixed now.

Issue #2080, opened 2018-11-26
Support to create C_Location via cypress


metas-ts:
### Is this a bug or feature request?
FR (might be bug too)

### What is the current behavior?

See the comment in `purchase_create_vendor_spec.js` in commit https://github.com/metasfresh/metasfresh-webui-frontend/commit/290b9005784ab3777a34e63cbe4aabef41833c8d

#### Which are the steps to reproduce?

run `purchase_create_vendor_spec.js` in cypress

### What is the expected or desired behavior?

should work

metas-ts:
imo no IT neccesary

PR #2089, opened 2018-11-29
Cypress tests to create C_Location

siemiatj:
Fixed the `purchase_create_vendor_spec` test file:

- Still not sure why clicking the Address button caused such weird behavior, but it might be some Cypress quirk. 
- `writeIntoLookupField` command seems useless. I was able to achieve the same with the existing `writeIntoCompositeLookupField`
- tab names will have to be changed once the rollout to w101 happens
- even though it's not necessary I've used aliases to requests and custom events so that it'll be easier to copy that to other test cases in the future

Related to #2080


cypress/integration/purchase_create_vendor_spec.js
```diff
@@ -0,0 +1,74 @@
...
+    cy.selectTab('Window-123-222');
+    cy.pressAddNewButton();
+    cy.writeIntoStringField('Name', 'Address');
+
+    cy.server()
+    cy.route('POST', '/rest/api/address').as('postAddress')
+    
+    // TODO: extract into a command
+    // Kuba: I'm not sure if this wouldn't be too specific for this use case.
+    cy.get('.form-field-C_Location_ID').find('button').click();
```
metas-ts: > I'm not sure if this wouldn't be too specific for this use case.

I think I don't understand.
We have these location widgets in many places, only the filedname (here: `C_Location_ID`) might vary.
And of course the
```JS
 cy.writeIntoStringField('City', 'Cologne')
        .writeIntoCompositeLookupField('C_Country_ID', 'Deu', 'Deutschland');
```
will vary.
I guess in JS we can do closures, right?
Maybe for a test-dev-user it might look like

```JS
cy.editAddressField(
	'C_Location_ID',
	function () {
	    cy.writeIntoStringField('City', 'Cologne')
        cy.writeIntoCompositeLookupField('C_Country_ID', 'Deu', 'Deutschland');
  	}
)
```
(i hope this makes sense and I didn't screw up the syntax!)
And the command would perform all the cy.server, route, emit etc.
wdyt?

Note: we have a similar thing for `M_AttributeSetInstance_ID`, e.g. in order lines. maybe we can even have one command for both cases? but i would also be fine with two commands (and if it's extra efffort, let's do `M_AttributeSetInstance_ID` later..but it will come soon)

cypress/support/commands.js
```diff
@@ -299,6 +301,15 @@ Cypress.Commands.add('waitForHeader', (pageName, breadcrumbNr) => {
   });
 });
 
+Cypress.Commands.add('selectSingleTabRow', (tabName) => {
```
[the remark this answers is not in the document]
metas-ts: @siemiatj codacy is right, my bad; can you please remove 
`tabName` (or if you can think of something useful to do with the parameter, then do that :-))

cypress/integration/purchase_create_vendor_spec.js
siemiatj: We're talking two different things here. What I meant is having a custom command for clicking a button. It saves us maybe 20 characters tops when used multiple times. And sometimes we might want to check this button's text for instance. So having this command makes no sense to me and as far as I understood, it's against the recommended best practices.

The rest is fine.

siemiatj:
Should be ready to merge now.

Issue #1978, opened 2018-09-25
Make actions cypress friendly


metas-ts:
### Is this a bug or feature request?
FR
analog to https://github.com/metasfresh/metasfresh-webui-frontend/issues/1926

The concept was described in metasfresh/metasfresh-webui-api#1023

See as of https://github.com/metasfresh/metasfresh-webui-api/issues/1055
"actions" now also have an internal name.
Please do something similar analog to #1926

metas-dh:
Results of IT1
tested in mf15

* cypress test for quick actions using internal name works fine, see #1995: OK
* cypress test for JSONDocumentAction using internal name does not yet work: 
=> checked with metas-ts: this was also in the scope of this task, see https://github.com/metasfresh/metasfresh-webui-api/issues/1055
@siemiatj pls make document actions cypress friendly as well

metas-ts:
@siemiatj 

i.e. when i click on the wise 3 dots `meta-icon-more` :
![image](https://user-images.githubusercontent.com/1264789/49657243-82d9de00-fa3f-11e8-9a5f-65095bd6d16b.png)

the API returns (in this case) the following:
```JSON
{
    "actions": [
        {
            "processId": "ADP_531089",
            "caption": "Auftrag aus Angebot",
            "description": "Erzeugt einen Auftrag aus einem bestehenden Angebot",
            "quickAction": false,
            "defaultQuickAction": false,
            "evaluateDuration": "19.60 μs",
            "internalName": "C_Order CreateNewFromProposal"
        },
        {
            "processId": "ADP_540602",
            "caption": "Bestellkontrolle zum Auftrag eigener Ausdruck",
            "description": "Druckt den Zusatz zur Auftragsbestätigung, ohne das betreffende Druckstück einer bestimmten Person zuzuordnen. Dadurch wird das Druckstück letztlich der Person zugeordnet, die diesen Prozess ausführt.",
            "quickAction": false,
            "defaultQuickAction": false,
            "evaluateDuration": "25.80 μs",
            "internalName": "C_Order_OrderCheckup_Jasper_NoRouting"
        }
    ]
}
```
..and we would like to have the `internalName` be embedded in the page,
and a command such that we can do

```js
cy.executeProcess('C_Order_OrderCheckup_Jasper_NoRouting');
```

..and cypress will fire the respective process

metas-ts:
works; used when creating new contracts for bpartners

PR #2099, opened 2018-12-08
Improve dealing with actions in cypress

siemiatj:
Related to #1978


metas-ts:
I have https://github.com/metasfresh/metasfresh-webui-frontend/commit/a21c246 which is the latest commit of this PR
image

The Id with the internal names are there
image

PR #2152, opened 2019-01-17
Add description to table cells

siemiatj:
Related to #2137


siemiatj:
Added description to all widgets as discussed with @metas-ts . Ready to review.

metas-ts:
### checking it locally
#### Reference to record with description
image

#### Ref-List I:
image

#### Ref-List II:
image


#### Table/Grid cell
image

PR #2166, opened 2019-01-30
[WiP] Collapsible form groups

siemiatj:
Waiting for flag to be added in the API.

collapse


siemiatj:
We're not doing groups but sections, and this is already handled by the single entry PR.

Issue #2169, opened 2019-02-04
Host google fonts locally


siemiatj:
### Is this a bug or feature request?

Feature

### What is the current behavior?

Roboto font is loaded from google fonts.

#### Which are the steps to reproduce?

.

### What is the expected or desired behavior?

We should load google fonts locally.

metas-ts:
"Dev-Test": https://github.com/metasfresh/metasfresh-webui-frontend/pull/2170#issuecomment-460206821

metas-dh:
Results of IT1
tested in mf15

Fonts are hosted locally: OK

PR #2170, opened 2019-02-04
Load google fonts from local

siemiatj:
Related to #2169


metas-ts:
# Before this change

----------------------

image


-----------------------

# After this change

image

PR #2181, opened 2019-02-13
Fix modal error when creating process

siemiatj:
I mistakenly used immutable selector on a string.

Related to #2157 and this [comment](https://github.com/metasfresh/metasfresh-webui-frontend/issues/2157#issuecomment-463200589)


siemiatj:
See the comment in the issue : https://github.com/metasfresh/metasfresh-webui-frontend/issues/2157#issuecomment-463274879

Issue #2204, opened 2019-03-06
QR Code not displayed


siemiatj:
### Is this a bug or feature request?

Bug

### What is the current behavior?

QR code is not rendered in the modal.

![screen shot 2019-03-06 at 16 50 46](https://user-images.githubusercontent.com/513460/53894928-3fd4da00-4031-11e9-807c-58776bab4536.png)

#### Which are the steps to reproduce?

For instance run `4 eyes review` action or any other that should show a qr code.

### What is the expected or desired behavior?

![screen shot 2019-03-06 at 17 00 28](https://user-images.githubusercontent.com/513460/53894910-39def900-4031-11e9-9fe2-f05f5991b02b.png)

metas-ts:
🍒 picked the 4 commits of PR https://github.com/metasfresh/metasfresh-webui-frontend/pull/2205 to `release`

PR #2205, opened 2019-03-06
QR Code fixes

siemiatj:
Related to #2204


siemiatj:
Demo instance on dev328. Works locally.

Issue #2177, opened 2019-02-11
Support single entry view mode for Tabs


siemiatj:
### Is this a bug or feature request?

Feature

### What is the current behavior?

Right now tabs will always display items in tables.

#### Which are the steps to reproduce?

Open any tab with a single entry in the table.

### What is the expected or desired behavior?

For single entries hidden behind a new flag we want tabs to support showing records in a detailed view, without the table and `add new record` button.

metas-ts:
Hi @siemiatj 
I'm in the process of providing you with a def instance.
Here is what you will get from the API:
![image](https://user-images.githubusercontent.com/1264789/52663641-6064b500-2f07-11e9-8a9e-fffc0307b684.png)

Notes this tab
* has `singleRowDetailView`
  * for "normal" tabs (`singleRowDetailView==false`) the API omits the property altogether; let me know if it should harather be explicitly provided by the API in both cases
* does not have a list of `elements` (let me know if you instead want an explicitly empty list)
* has a list of `sections`, just like the window's main part

metas-ts:
this close "by me" was accidental, by me merging https://github.com/metasfresh/metasfresh-webui-frontend/commit/93afbcb5fa9fae42e631e36d6059e95921a72dd7

Issue #2160, opened 2019-01-25
Selected row not always correctly recognized


siemiatj:
### Is this a bug or feature request?
Bug
### What is the current behavior?
Sometimes in details view row is selected, but header doesn't react to that and row-specific actions are still disabled.

![screen shot 2019-01-25 at 01 40 06](https://user-images.githubusercontent.com/513460/51718194-7d167700-2044-11e9-8fd4-8cd47ff19837.png)

#### Which are the steps to reproduce?

Go to sales order and refresh the view clearing cache.

### What is the expected or desired behavior?

After selecting rows, actions should be immediately available.

metas-dh:
Results of IT1
tested in mf15

* actions immediately available after selecting rows (e.g. in sales order): OK
* calling an action for a selected line: nothing happened for me, e.g. `Create purchase orders` (SetUp was OK): nothing happening, just an error in console:

```
numeral.js:11 Uncaught (in promise) TypeError: _.last is not a function
    at Modal.<anonymous> (numeral.js:11)
    at w (react.production.min.js:10)
    at Generator._invoke (react.production.min.js:10)
    at Generator.e.(anonymous function) [as next] (https://mf15cloudit/bundle-3d67939f7264a1d2f26f-git-87c153d.js:149:11411)
    at asyncGeneratorStep (numeral.js:11)
    at a (numeral.js:11)
    at numeral.js:11
    at new Promise (<anonymous>)
    at Modal.init (numeral.js:11)
    at Modal.<anonymous> (numeral.js:11)

```

* other problem: i had all actions available also when there was no row selected, e.g. in sales order, no orderline was selected and i could select the `Multigroup lines` action: => nothing happens, same error in console

siemiatj:
This can be re-tested

metas-dh:
Results of IT2
tested in mf15

tested with sales order:
* actions immediately available after selecting rows: OK
* calling an action for a selected line: correct modal opens (e.g. Create purchase order, or Pricing Conditions): OK
* only the suitable actions are available for the selection:
   * 1 line selected: all actions (in sales order) for orderline available: OK
   * >1 line selected: correct actions available (e.g. Create purchase order, Grouping, but not Pricing Conditions): OK
   * no line selected: correct actions available (e.g. Purchase order from sales order, but not  Pricing Conditions): OK

PR #2206, opened 2019-03-07
About section changes and single entry changes

siemiatj:
Related to #1945 & #2177 

Changes related to the About section:
- using translated strings
- showing About info for lists of documents as well
- on lists, not showing info when > 1 element selected
- in details view , when > 1 child element selected, not showing Row info

Details view:
- styling of sections
- loading collapsed/uncollapsed state
- sections title clickable

General fixes:
- change how tabs are fetched after closing the modal
- fix deselecting elements on the lists


siemiatj:
Rolled out to dev330. Also related to #2145 & #2160 as this fixes old selection bug that went unnoticed.

siemiatj:
Checked locally and looks ok.

PR #2227, opened 2019-03-26
Force our custom time formats

siemiatj:
Related to #2221

### DIFF
--- a/src/components/widget/DatePicker.js
+++ b/src/components/widget/DatePicker.js
@@ -109,6 +109,8 @@ class DatePicker extends Component {
   );
 
   render() {
+    const format = this.state.open ? true : this.props.timeFormat;
+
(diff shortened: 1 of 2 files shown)


teosarca:
Did a quick test. See my comments bellow.
* grid view: make sure the "seconds" are not displayed
* single document layout (widget): make sure the "seconds" are not displayed
* super IMPORTANT for our customers: If the Locale is using 24h time format, we have to respect that. i.e. switch to German language and try it.

To summarize: all that we want is to cut out the seconds from everywhere.

image

image

metas-ts:
Reverted me integrating this PR as of teo's review https://github.com/metasfresh/metasfresh-webui-frontend/pull/2227#issuecomment-476780896

Issue #2230, opened 2019-03-27
Refresh filtered list


siemiatj:
### Is this a bug or feature request?

Bug

### What is the current behavior?

Right now when an element on a filtered list is edited in place (for instance by calling an action) and does not pass the filter check it's still visible to the user.

#### Which are the steps to reproduce?

Go to Phonecall Schedule and activate filter by `isCalled` to false.
Select a filtered element and run a header action `Mark As Called`.

### What is the expected or desired behavior?

Element should be removed from the list.

PR #2231, opened 2019-03-27
Force filtering list when element changed in-place

siemiatj:
Related to #2230


siemiatj:
Note to IT:

I tried describing how to repro in the ticket description.

Issue #2241, opened 2019-04-04
Order line quick input not working


metas-ts:
### Is this a bug or feature request?
Bug

### What is the current behavior?

the API expects a POST to /rest/api/window/..../complete 
which does not happen on dev393. when i hit end and the value was changed, then a PATCH is done...that's all
i didn't get to try out other instances though. only https://dev393.metasfresh.com
(user ts pw demo1234)
still i think this would be a show-stopper for next week's release. going to veryfy it'S happing also on the current master, then going to create a ticket for you

metas-ts:
Sidenote, may or may not be related: i also can't increment and decrement the number with these thingies anymore: 
![image](https://user-images.githubusercontent.com/1264789/55554276-ac9bcc80-56e2-11e9-9e35-866f2492247a.png)

metas-ts:
🍒 picked https://github.com/metasfresh/metasfresh-webui-frontend/issues/2241 to `release` branch

metas-dh:
Results of IT1
tested in release, during e2e

* quick input working fine again, orderline is created correctly after hitting `Enter`: **OK**
* input with `Add new` also working: **OK**
* incrementing and decrementing the number in quick input using up/down arrows in qty field working fine as well: **OK**

PR #2242, opened 2019-04-04
Fix submitting QuickInput form

siemiatj:
Remove some unnecessary old code as well #2241


siemiatj:
For some reason I haven't pushed one commit. Thanks god for the git-reflog.

PR #2250, opened 2019-04-23
Rewrite table navigation to resemble Google Sheets/Excell

siemiatj:
Related to #2156 

REQUIREMENTS:
* [x] fields with `viewEditorRenderMode` set to `always` should be in a constant edit mode, showing input at all times
* [ ] fields with value set to `onDemand` should turn editable after `double-click`/`{enter}` key. (*haven't tested that*)
* [x] field that is editable is immediately patched after pressing `{enter}/{tab}` 
* [x] field that is editable is immediately patched after mouse-clicking outside
    * [x] text input
    * [x] textarea
    * [x] number
* [x] table navigation with `Arrows` and `{tab}` keys
* When cell is focused:
  * Focus fields
    * [x] text input
    * [x] textarea
    * [x] number
    * [x] lookup
  * [x] focus attributes button, but don't open it
    * [x] open attributes panel after pressing `{enter}`
  * [x] focus list input, but don't open the dropdown
    * [x] open dropdown after pressing `{down}` (so this is seamless with what we normally have for widgets)
  * [x] `Arrows` don't traverse between cells but navigate text/lists
  * [x] `{tab}` still switches to next cell
    * [x] tabbing out submits the input
        * [x] number
        * [x] text
        * [x] textarea
    * [x] tabbing out closes list/lookup dropdown
    * [x] tabbing out closes attributes modal
  * [x] `{enter}` submits the field and blurs the widget, but cell is still selected
* if cell is selected, but widget blurred (after submit)
  * [x] `{enter}` re-focuses the widget
    * [x] number
    * [x] text
    * [x] textarea
    * [x] list
    * [x] lookup
  * [x] `{arrows}/{tab}` work for navigating between the cells (as normal)
  * [x] navigation inside attributes dropdown works
  * [x] `{tab}` in last input of attributes dropdown closes the dropdown
  * [x] location attributes dropdown works correctly

TODO:
* [ ] TESTS (#2260)


ionut-movila:
@siemiatj And for attributes we would need little help to figure out how key handling propagation works. Thank you.

siemiatj:
@teosarca apart of the attributes modal that doesn't hide on tab, the rest is good for checking

teosarca:
First issue: quick actions are not fetched for modal views
e.g.
* receipt candidates: http://localhost:3000/window/540196
* select one and press on quick actions to receive
* the modal HU editor opens
* click around on rows => /quickActions is no longer called

image

siemiatj:
It is called for me http://recordit.co/1t4VRSVaPr . It won't be called when traversing horizontally as the selected row doesn't change.

teosarca:
hmm, now i cannot reproduce neither.... but now I got another one...

teosarca:
On a new window, i got following error in console. Any idea why and if u can blindly fix it?
(note: on orders seems to work)

image

teosarca:
found another thing: if ur user has multiple roles assigned, now it's hard to login without using the mouse.
i.e. when u have to pick the role, ENTER just opens the downdown popup:

image


@metas-mk  do u think it's important?

teosarca:
also, when I open HU editor from receipt schedules, i get some warnings in console. Pls let's get rid of them.

image

image

teosarca:
@siemiatj  quickly checked and I have one requirement more to make it usable to the data entry users.

Take a look at the screenshot below.
(the requirements apply only for modals!)
* can we take out the "up-down" controls from numeric fields
* make sure that pressing "up" and "down" keyboard arrows will move the cell to up/down
* any other suggestions to make it fluent for a user?


image

teosarca:
@siemiatj  for now integrating as is

Issue #2262, opened 2019-05-14
Scrollbar missing in modal overlays


metas-dh:
### Is this a bug or feature request?
Bug
### What is the current behavior?
Scrollbar is missing in modal overlays, on the right, so you cannot scroll down in attributes / to see all HUs when there are more than fit the initial view.

material receipt cands:

![2019-05-14 09_18_16-Metasfresh](https://user-images.githubusercontent.com/18696715/57678589-d7563a80-7629-11e9-881f-4819d0083773.png)

manufacturing order:

![2019-05-14 09_18_50-Metasfresh](https://user-images.githubusercontent.com/18696715/57678598-dc1aee80-7629-11e9-954e-9280fe45a0f1.png)

#### Which are the steps to reproduce?
Try and see.
### What is the expected or desired behavior?
Show the scrollbar.

metas-Kay:
also missing in product proposal overlay:
![scroll bar missing](https://user-images.githubusercontent.com/29163573/57772780-a43da500-7716-11e9-9e5c-e4f98a0e6397.png)

metas-dh:
Results of IT1
tested in release

Scrollbar displayed correctly, in e.g. material receipt cands, manufacturing order, product proposal etc.: **OK**

PR #2266, opened 2019-05-15
Issue #2262 fixed Scrollbar missing in modal overlays

oliviupis:
Related to #2262

### DIFF
--- a/src/components/app/DocumentList.js
+++ b/src/components/app/DocumentList.js
@@ -969,7 +969,8 @@ export class DocumentList extends Component {
               tabIndex={0}
               indentSupported={layout.supportTree}
               disableOnClickOutside={clickOutsideLock}
-              limitOnClickOutside={isIncluded && isModal}
+              // limitOnClickOutside={isIncluded && isModal}
+              limitOnClickOutside={isModal}
               defaultSelected={selected}
               refreshSelection={refreshSelection}
               queryLimitHit={data.queryLimitHit}
--- a/src/components/app/DocumentList.js
+++ b/src/components/app/DocumentList.js
@@ -969,7 +969,6 @@ export class DocumentList extends Component {
               tabIndex={0}
               indentSupported={layout.supportTree}
               disableOnClickOutside={clickOutsideLock}
-              // limitOnClickOutside={isIncluded && isModal}
               limitOnClickOutside={isModal}
               defaultSelected={selected}
               refreshSelection={refreshSelection}
--- a/src/components/table/Table.js
+++ b/src/components/table/Table.js
@@ -475,7 +475,8 @@ class Table extends Component {
     const parentNode = event.target.parentNode;
     const closeIncluded =
       limitOnClickOutside &&
-      parentNode.className.includes('document-list-wrapper')
+      (parentNode.className.includes('document-list-wrapper') ||
+        event.target.className.includes('document-list-wrapper'))
         ? parentNode.className.includes('document-list-has-included')
         : true;
 
@@ -1312,9 +1313,13 @@ const mapStateToProps = state => ({
   allowOutsideClick: state.windowHandler.allowOutsideClick,
 });
 
+const clickOutsideConfig = {
+  excludeScrollbar: true,
+};
+
 export default connect(
   mapStateToProps,
   false,
   false,
   { withRef: true }
-)(onClickOutside(Table));
+)(onClickOutside(Table, clickOutsideConfig));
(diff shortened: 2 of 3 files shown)


teosarca:
issue https://github.com/metasfresh/metasfresh-webui-frontend/issues/2262

teosarca:
Quickly checked but had some issues.

Test case:
* open Material receipt candidates: http://localhost:3000/window/540196
image

* select a row (e.g. first one) and click on the action button (u might have slightly different numbers in ur case but don't bother) => the HU editor modal will be opened
image
=> HU editor:
image


* Remark: we have a scroll vertical scroll for the whole modal panel (OK)
* try to click on it in order to move it up-down => the scroll disappears because:
   * first row from the left table is selected 
   * if the first row is deselected then there are no attributes to be shown for that row
   * if there are no attributes to be shown then the layout fits into that are so there is no need for vertical scrolling

Expectation: when i try to use the scroll up-down the row from the left table shall not get deselected.
NOTE: if this is a bigger topic and not just a quick/hot fix pls let me know.

image

image

ionut-movila:
@teosarca We did investigate the problem here and look like the table row gets deselected due to lib. onClickOutside, that is used as a HOC wrapper with Table component. And because the scroll where we put it in DOM is "outside" of the Table it triggers this behaviour of deselecting the row.

I'm attaching here a highlight of the area with the Table, so it's easy to see what is considered "outside". IMO this is very UX confusing, and I suggest row deselection should happen with a click on selected row, like "click to select and click one more time to deselect". What do you think?
Click-outside-area

siemiatj:
@ionut-movila unfortunately deselects have to be trigger on everything outside of table body. It's not great, it adds a lot of hacking but apparently that's what's considered natural for our clients. `onClickOutside` has the option to not listen to clicks on scrollbars, but it doesn't always work correctly as far as I remember: https://www.npmjs.com/package/react-onclickoutside#regulate-whether-or-not-to-listen-to-scrollbar-clicks

siemiatj:
I forgot I've already introduced a way of limiting when `onClickOutside` will react to events. Pushed this to a temp branch:

https://github.com/metasfresh/metasfresh-webui-frontend/commit/4b64045d785041bb266d54cf0413a5986f319987

and here's the config flag that controls it:

https://github.com/metasfresh/metasfresh-webui-frontend/blob/master/src/components/app/DocumentList.js#L972

src/components/table/Table.js
```diff
@@ -475,7 +475,8 @@ class Table extends Component {
     const parentNode = event.target.parentNode;
     const closeIncluded =
       limitOnClickOutside &&
-      parentNode.className.includes('document-list-wrapper')
+      (parentNode.className.includes('document-list-wrapper') ||
+        event.target.className.includes('document-list-wrapper'))
```
ionut-movila: This does help, thanks!

src/components/app/DocumentList.js
```diff
@@ -969,7 +969,8 @@ export class DocumentList extends Component {
               tabIndex={0}
               indentSupported={layout.supportTree}
               disableOnClickOutside={clickOutsideLock}
-              limitOnClickOutside={isIncluded && isModal}
+              // limitOnClickOutside={isIncluded && isModal}
+              limitOnClickOutside={isModal}
```
ionut-movila: @siemiatj: Not sure if this is the correct way, but it looks like prop. isIncluded is undefined for DocumentList when used inside a RawModal. This happens because  in Container component, we use <DocumentList> at line 154 with no isIncluded prop. Is this correct?

src/components/table/Table.js
```diff
@@ -1312,9 +1313,13 @@ const mapStateToProps = state => ({
   allowOutsideClick: state.windowHandler.allowOutsideClick,
 });
 
+const clickOutsideConfig = {
+  excludeScrollbar: true,
```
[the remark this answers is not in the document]
siemiatj: oh interesting, I missed this part.

src/components/app/DocumentList.js
siemiatj: It is set when there are two grids in a RawModal. But maybe this `isIncluded` can be skipped here, or we can create another case ?

Screen Shot 2019-05-21 at 17 20 18

Did a quick recording so that you can check this use case http://recordit.co/QXHEyQdV6I

ionut-movila: Even for this case the solution to have `limitOnClickOutside={isModal}` is working. So I just removed the commented  old version and this should be ready to merge.

PR #2269, opened 2019-05-16
Issue 2268 action menu is now correctly positioned

oliviupis:
Related to #2268


siemiatj:
Can you fix it for mobile views too ?

met1


met2

PR #2273, opened 2019-05-20
FR #2223 headerProperties added

oliviupis:
Related to #2223

### DIFF
--- a/src/assets/css/panel.css
+++ b/src/assets/css/panel.css
@@ -114,6 +114,18 @@
     padding: 0 2rem;
     box-shadow: none;
     transition: all 0.2s ease;
+
+    .optional {
+        flex: 9;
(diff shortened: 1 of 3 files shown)


siemiatj:
Works ok. But I'm not sure about styling. I imagined something more along those lines:

Screen Shot 2019-05-21 at 14 17 55

instead of current

Screen Shot 2019-05-21 at 14 19 16

@teosarca wdyt ?

teosarca:
@siemiatj  Fully agree. This is how I thought it initially.

Additionally to the picture u provided i would like to say:
* the names shall be **bold**
* after the semicolon (i.e. ":") we shall have a space

e.g.
**Order Document No:** 12345

siemiatj:
Merge conflict

Issue #2274, opened 2019-05-21
Calendar icon should react to clicks


siemiatj:
### Is this a bug or feature request?

Bug

### What is the current behavior?

From @TheBestPessimist :

> Problem: on a date field/row, i can click the date field (be it empty or already showing a date) to show the calendar date picker, but i cannot click the calendar icon to the right to do the same.

#### Which are the steps to reproduce?

![qehqA7QEMf](https://user-images.githubusercontent.com/4482210/58015341-a67f7500-7b03-11e9-8531-3509685fbf2f.gif)

PR #2277, opened 2019-05-22
Dev #2274 repositioning calendar and icon press works

oliviupis:
Related to #2274


oliviupis:
I've found another issue after commit and I'll work on this to make it better in all cases.

PR #2283, opened 2019-05-28
Dev 1740 email editor has address lookup behaviour

oliviupis:
Related to #1740


oliviupis:
It doesn't need event handlers while navigate through dropdown list, only on selection it requires an event of changing state.

Issue #2149, opened 2019-01-16
Decimal value .00 not added to product price at first


metas-dh:
### Is this a bug or feature request?
Bug
### What is the current behavior?
When you add a product price and just set e.g. 2, the decimal value .00 is not added at first. After closing the modal overlay, the .00 is added to the price.

When setting the price:
![2019-01-16 14_00_05-notificationsform](https://user-images.githubusercontent.com/18696715/51250695-7f6b3800-1997-11e9-8545-e1afc172e2c8.png)

After closing and reopening the overlay:
![2019-01-16 14_02_26-notificationsform](https://user-images.githubusercontent.com/18696715/51250702-84c88280-1997-11e9-9e14-0cfde6e87a80.png)

#### Which are the steps to reproduce?
Try and see.
### What is the expected or desired behavior?
Decimal value shall be added directly after setting the price.

(not sure if this is only a frontend issue)

siemiatj:
@oliviupis let me know in case it turns out to be a backend thing

metas-dh:
IT successful, see me03 issue.

PR #2290, opened 2019-06-05
#2292-me, fixing mobile double click for table rows.

ionut-movila:
This should fix the double click on mobile, but still due to missing pagination it will be working very slow (not really usable).

I have used a new lib to handle mobile double tap: https://github.com/JedWatson/react-hammerjs, because with a simple test when running on mobile view in browser `onDoubleClick` will not fire, see http://fiddle.jshell.net/jxg/z2hdqw7L/show/


teosarca:
task: https://github.com/metasfresh/metasfresh-webui-frontend/issues/2292

src/components/table/Table.js
```diff
@@ -1301,6 +1305,13 @@ class Table extends Component {
             handleToggleExpand={toggleFullScreen}
           />
         )}
+        {isMobileOrTablet && rows.length > MOBILE_TABLE_SIZE_LIMIT && (
+          <span className="text-danger">
+            {/*TODO: use here translations, but how we define new keys?*/}
```
siemiatj: @ionut-movila ping @teosarca or @metas-ts and ask them to add whatever translations you need

src/components/table/Table.js
```diff
@@ -967,93 +969,95 @@ class Table extends Component {
...
-          return collapsedRows.indexOf(row[keyProperty]) === -1;
+    let renderRows = rows.filter(row => {
+      if (collapsedRows) {
+        return collapsedRows.indexOf(row[keyProperty]) === -1;
+      }
+      return true;
+    });
+
+    if (isMobileOrTablet && rows.length > MOBILE_TABLE_SIZE_LIMIT) {
+      renderRows = renderRows.slice(0, MOBILE_TABLE_SIZE_LIMIT);
```
siemiatj: So how do you switch to the next page if pagination is hidden on mobile devices ?

teosarca: my suggestion, as a quick&dirty solution, was to limit the query to a reasonable number of rows.
If that amount is hit, we display the "limit" message at the bottom of the page.
User has to filter and search if he needs more.
In this way we make the window usable with small amount of effort.

----
THE solution: live scroll or some pagination which looks nice on mobile... but I think live scroll is THE shit!

src/components/table/Table.js
teosarca: introduced `view.limitTo`.
See https://w101.metasfresh.com:8443/rest/api/i18n/messages?filter=view.limitTo for current value.

src/components/table/Table.js
siemiatj: Yeah, we have an issue to implement that. The only problem is when we're showing nested data.

siemiatj:
Just use the new translation and it's good to merge.

teosarca:

teosarca:
@ionut-movila  still have issues with this one. Let's check it together on monday.

src/components/table/Table.js
ionut-movila: I have fixed this with commit: https://github.com/metasfresh/metasfresh-webui-frontend/pull/2290/commits/ccd7caaf1b0d7470acb1c6a816fadd4ec411196f

@metas-ts But we need to update the string from backend with respect to interpolation rules of the FE lib: https://www.npmjs.com/package/counterpart#interpolation

So it should be like this: "Limited to %(limit)s rows of %(total)s"

teosarca: done

teosarca:
quickly checked but i got:
image

Issue #2299, opened 2019-06-10
 Filter view should recover when errors happen


TheBestPessimist:
### Is this a bug or feature request?
BUG

### What is the current behavior?

When filtering and an error is thrown, the error is displayed (✅) but the "loading screen" over the data remains forever (as if waiting for the response) (🚫).
The only way to recover from this is to reload the view, which destroys the filter input (which is bad).

![0000000145](https://user-images.githubusercontent.com/4482210/59176992-57a58800-8b63-11e9-8e67-601304e4bad1.gif)

#### Which are the steps to reproduce?

Example to test on: Business Partner window (`/window/123`), `Area Search` filter -> add 99999999999999 distance.

### What is the expected or desired behavior?

After the error is thrown, the "loading screen" should disappear and the filter view should be shown again, with all the user input preserved, so that the user can fix it and re-search.

PR #2301, opened 2019-06-11
Dev 2299

oliviupis:
#2299


oliviupis:
WIP changing current implementation, no review needed.

PR #2308, opened 2019-06-29
Dev cypress run hangs

metas-ts:
Thanks a lot to @astehlik ( https://github.com/cypress-io/cypress/issues/1235#issuecomment-411839157 ) !


metas-ts:
🍒 picked these 2 commits to branch `release`

Issue #2278, opened 2019-05-23
Fix decimal comma in Qty Fields


metas-mk:
### Is this a bug or feature request?
Bug
### What is the current behavior?
Currently, it does not work typing comma in decimal value numeric fields. Only period works, no matter which language is set.
#### Which are the steps to reproduce?
Open window /window/540325 and try to edit numeric value in price values with comma.
![image](https://user-images.githubusercontent.com/13365687/58257111-8bfe0380-7d70-11e9-8324-5d87eff76af4.png)
### What is the expected or desired behavior?
Shall work with comma and period.

ionut-movila:
@metas-mk Can you please check the problem one more time? Looks to be fixed when we tested today running from master branch. See [here](http://recordit.co/75Kz3wSOXK) video.

metas-mk:
@ionut-movila can u pls always show the url bar in pics/ videos so I can check,

oliviupis:
@metas-mk .../window/540325/540025

metas-mk:
@oliviupis here a screenshot of what I see on https://w101.metasfresh.com:8443/window/540325/540025

![image](https://user-images.githubusercontent.com/13365687/58869313-44577000-86be-11e9-824e-4f0c0294d22b.png)

No comma allowed. Onlly decimal points.

PR #2316, opened 2019-07-12
Support both comma and dot as number precision separators

siemiatj:
Related to #2278


siemiatj:
Ready to review. I'm adding additional tests in the meantime.

src/components/widget/CustomNumberInput.js
```diff
@@ -0,0 +1,53 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+
+const FLOAT_NUMBER_REGEX_FORMAT = /^-?[0-9]+[.,]?[0-9]*$/;
```
metas-ts: I think people will want to write e.g. `,1` for `0.1`...
maybe...if it's easy...and you now even have a test to cover your back...
could you look into that pls 😊 ?

src/components/widget/RawWidget.js
```diff
@@ -129,9 +130,16 @@ export class RawWidget extends Component {
...
   // and send a patch request only if date is changed
   handlePatch = (property, value, id, valueTo, isForce) => {
-    const { handlePatch } = this.props;
+    const { handlePatch, widgetData } = this.props;
     const willPatch = this.willPatch(property, value, valueTo);
 
+    const fieldData = widgetData.find(widget => widget.field === property);
+    let isCostPriceInput = false;
+
+    if (fieldData && fieldData.widgetType === 'CostPrice') {
```
metas-ts: do we still need this? AFAIU your PR is about a generic solution..?

siemiatj: I don't want to go so far and turn this for all fields. It's gonna be madness, I'm pretty sure.

src/components/widget/CustomNumberInput.js
siemiatj: We've never supportad that but I might as well just add that.

PR #2319, opened 2019-07-17
Fix filters errors

siemiatj:
Fixes error when applying/clearing filters. Also fixes error when requesting Attributes. Both of those issues where caused by fields that are not (?) required by the backend anymore. 

Related to #2318

### DIFF
--- a/src/api/view.js
+++ b/src/api/view.js
@@ -75,8 +75,7 @@ export function createViewRequest({
 
 export function filterViewRequest(windowId, viewId, filters) {
   filters.map(filter => {
-
-    delete filter.defaultVal
+    delete filter.defaultVal;
 
     filter.parameters &&
       filter.parameters.map((param, index) => {
--- a/src/api/view.js
+++ b/src/api/view.js
@@ -75,8 +75,15 @@ export function createViewRequest({
 
 export function filterViewRequest(windowId, viewId, filters) {
   filters.map(filter => {
+
+    delete filter.defaultVal
+
     filter.parameters &&
       filter.parameters.map((param, index) => {
+        // @TODO: This is really ugly
+        delete param.defaultValue;
+        delete param.defaultValueTo;
+
         if (param.widgetType === 'Date' && param.value) {
           filter.parameters[index].value = Moment(param.value).format(
             DATE_FORMAT
--- a/src/api/view.js
+++ b/src/api/view.js
@@ -1,7 +1,5 @@
 import { post, get, delete as del } from 'axios';
-import Moment from 'moment';
-import { getQueryString } from '../utils';
-import { DATE_FORMAT } from '../constants/Constants';
+import { getQueryString, cleanupFilter } from '../utils';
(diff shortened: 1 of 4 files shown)


src/api/view.js
```diff
@@ -75,8 +75,14 @@ export function createViewRequest({
 
 export function filterViewRequest(windowId, viewId, filters) {
   filters.map(filter => {
+    delete filter.defaultVal;
+
     filter.parameters &&
       filter.parameters.map((param, index) => {
+        // @TODO: This is really ugly
+        delete param.defaultValue;
+        delete param.defaultValueTo;
```
teosarca: basically, the param object shall contain only following 3 properties:
* parameterName
* value
* valueTo

Why not building an object only with those instead of manipulating some other object copied from hells knows where?

src/api/view.js
```diff
@@ -75,8 +75,14 @@ export function createViewRequest({
 
 export function filterViewRequest(windowId, viewId, filters) {
   filters.map(filter => {
+    delete filter.defaultVal;
```
teosarca: the filter shall contain only following 2 params:
* `filterId`
* `parameters`

why not building an object only with those 2?

src/api/window.js
```diff
@@ -18,27 +18,9 @@ export function initLayout(
...
-  docType,
-  docId,
-  tabId,
-  rowId,
-  fieldName,
-  entity
-) {
-  const type = entity === 'process' ? 'processId' : 'windowId';
-
+export function getAttributesInstance(attrType, tmpId) {
```
teosarca: hmmm, have to check this.... 
I mean, it would work for /address endpoint but not for /attributes endpoint.

On server side we are handling them differently but on client side it seems they are handle it the same way...
Will check and update here.

src/api/view.js
siemiatj: Is this documented anywhere ? I don't want to guess this kind of stuff.

src/api/view.js
siemiatj: We're getting filter params from the backend with all of this, so I guess this is what you mean by `hell` here.

src/api/view.js
teosarca: sure it is, swagger is our source of true.
In this case: http://w101.metasfresh.com:8080/swagger-ui.html#/view-rest-controller/filterViewUsingPOST

image



The only this is not obvious on swagger is that u don't have to set `caption` field.

src/api/view.js
teosarca: the server won't be sad if frontend is not returning all of those properties back :)))

src/api/window.js
teosarca: update: introduced the `source` for /address endpoint too (commit https://github.com/metasfresh/metasfresh-webui-api/commit/266ededcb9ed38fec05d43a39092d8d30979495d).

So, pls revert this and allow sending the source too.

PR #2321, opened 2019-07-18
Revert back to normal number field

siemiatj:
Related to #2278

I'm leaving the custom field for future reference.


siemiatj:
Because the presentation table cells are only showing the data from the server, to format them properly we need to use the application language. If it's set to `de_<whatever>` everything works fine. Inputs formatting is controlled by the browser.

Issue #2325, opened 2019-07-25
frontend patches too much in quick input field


metas-ts:
### Is this a bug or feature request?
Bug

This bug blocks
* https://github.com/metasfresh/metasfresh-e2e/issues/153
* https://github.com/metasfresh/metasfresh-e2e/issues/213
* *very* probably also affects further cypress specs

#### Quickentry in cypress tests 

* almost never works when executed on @TheBestPessimist's machine
* generally works when @TheBestPessimist does manual quickentry on his local machine
* generally works when I run both the cypress test and metasfresh locally
* almost never works when the cypress test is run locally against metasfresh running on dev618

There is one "wrong" empty string patch and we don't know where it's coming from:

#### Screenshot

<details>

![image](https://user-images.githubusercontent.com/1264789/61876909-34a21e00-aeee-11e9-9c9c-6ac4d787cbe4.png)

</details>

#### Screenshot-related notes:

payload of 1st "35" PATCH (OK)
`[{op: "replace", path: "Qty", value: "222"}]`

payload of 2nd "35" PATCH payload (NOK)
`[{op: "replace", path: "Qty", value: ""}]`

The following "complete" patch then fails because of the preceeding `value: ""` PATCH payload

#### Also note that

* the problem also happened when I commented out type `{selectall}, {force: true}` and instead did an "extra"
`cy.writeIntoStringField('Qty', '{enter}', false /*modal*/, undefined /*rewriteUrl*/, true /*noRequest*/);`
* even when it works locally on my machine, i see two `Qty` patches. The difference is that the 2nd one *also* has `value: "222"`

#### Which are the steps to reproduce?

* checkout the branch `gh153_ts`
* get the config.js for dev618: [config.js.zip](https://github.com/metasfresh/metasfresh-webui-frontend/files/3431706/config.js.zip)
* start the runner with `CYPRESS_baseUrl=https://dev618.metasfresh.com npm run cypress:open`
* you can copy this test to `cypress/integration/a`: [reverse_empties_return.js.zip](https://github.com/metasfresh/metasfresh-webui-frontend/files/3431704/reverse_empties_return.js.zip)
  * its hardcoded timestamp corresponds to a product, packaging, vendor etc etc which exsit on dev618, and it skips creating all those things

TheBestPessimist:
Also related: https://github.com/metasfresh/metasfresh-webui-frontend/issues/2322

metas-ts:
[...]

PR #2330, opened 2019-07-30
Make TableQuickInput widgets smarter

siemiatj:
Now when there's a submit to the server, fields won't patch by themselves.

I've run a bunch of cypress tests and everything works fine, so we should be good to go.

Related to #2325

### DIFF
--- a/src/components/widget/RawWidget.js
+++ b/src/components/widget/RawWidget.js
@@ -128,12 +128,12 @@ export class RawWidget extends Component {
   // Datepicker is checking the cached value in datepicker component itself
   // and send a patch request only if date is changed
   handlePatch = (property, value, id, valueTo, isForce) => {
-    const { handlePatch } = this.props;
+    const { handlePatch, inProgress } = this.props;
     const willPatch = this.willPatch(property, value, valueTo);
 
     // Do patch only when value is not equal state
     // or cache is set and it is not equal value
-    if ((isForce || willPatch) && handlePatch) {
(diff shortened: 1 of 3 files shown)


src/components/widget/RawWidget.js
```diff
@@ -128,12 +128,12 @@ export class RawWidget extends Component {
...
   // and send a patch request only if date is changed
   handlePatch = (property, value, id, valueTo, isForce) => {
-    const { handlePatch } = this.props;
+    const { handlePatch, inProgress } = this.props;
     const willPatch = this.willPatch(property, value, valueTo);
 
     // Do patch only when value is not equal state
     // or cache is set and it is not equal value
-    if ((isForce || willPatch) && handlePatch) {
+    if ((isForce || willPatch) && handlePatch && !inProgress) {
```
TheBestPessimist: since idk how frontend is supposed to work, i'm just throwing this here: if the patch response is so slow (think `Slow 3g mode`) and the user is editing the qty 3 times in a row really fast, is it possible that only the first edit (patch) is sent to backend and the other 2 are ignored, as `inProgress` did not turn to false fast enough?

siemiatj: `inProgress` is only set when you press `{enter}` to submit the quickinput. As a general rule of thumb, if you submit a form in a browser app - it's done, and you have to wait for the process to finish before doing any further changes. Until you submit it - you can change the values as many times as you want.

PR #2337, opened 2019-08-23
Support new date widget types

siemiatj:
Related to #2317


siemiatj:
Both issues fixed

siemiatj:
So I can't really reproduce this error with widget showing day earlier. I've tried switching system timezones, but still works ok. Fixed the other bug, simplified the whole thing a bit (as well as some super old TODO's) along the way. The last error is a bit tricky, and would require more digging. It doesn't really affect anything and is probably a combination of various things at the same time (the error React is showing in the console is misleading).

metas-mk:
API issue remaining. https://github.com/metasfresh/metasfresh-webui-api/issues/1249

Issue #2364, opened 2019-09-13
Organisation and client field misalligned 


metas-dh:
### Is this a bug or feature request?
EN
### What is the current behavior?
Atm the organisation and client field are misalligned in the bpartner window, now that org is a search field. In product window the allignment is fine.
(see pics in me03 issue)
#### Which are the steps to reproduce?
Try and see.
### What is the expected or desired behavior?
Shall be alligned in all windows.

PR #2379, opened 2019-09-23
Fix search input alignment

siemiatj:
Related to #2364


siemiatj:
Screen Shot 2019-09-23 at 20 02 29

siemiatj:
Yeah probably should've checked the BPartner window instead.

Screen Shot 2019-09-26 at 04 05 06


Screen Shot 2019-09-26 at 04 05 38

Issue #2370, opened 2019-09-18
Edited record is still visible after changing grid page


siemiatj:
### Is this a bug or feature request?

Bug

### What is the current behavior?

1. right click over number field => edit value
2. put any value
3. got to next page
4. NOK: record stays although we are on next page
![fsNMIsnIG9](https://user-images.githubusercontent.com/16079655/65042707-af6d3f80-d959-11e9-9aa5-fc32b9d9f728.gif)

#### Which are the steps to reproduce?

see behavior

### What is the expected or desired behavior?

Remove the edited record and use whatever comes with data

siemiatj:
The previous solution impacted the performance too much, so we need a better way to solve this

PR #2410, opened 2019-10-11
Fix refreshing rows when changing pages + performance improvements

siemiatj:
Apart of introducing a better solution to #2370, which doesn't impact the performance (at least) visibly this also cleans up and removes some unnecessary code complexity that caused thousands of re-renders that could be avoided. 

Here's an output from `why-did-you-update` performance measuring plugin (https://www.npmjs.com/package/why-did-you-update) for for table related components when selecting a row.

Before (almost 2000 redraws):

before

https://recordit.co/3VnnA3bx4Z

After (82, which is still not perfect but now it's Widget that is the bottleneck)

after

https://recordit.co/yE2TuYq8rU

### DIFF
--- a/package.json
+++ b/package.json
@@ -130,7 +130,7 @@
     "moment": "2.24.0",
     "moment-timezone": "0.5.26",
     "numeral": "2.0.6",
-    "object-hash": "^1.3.1",
+    "object-hash": "2.0.0",
     "prop-types": "15.7.2",
     "quagga": "0.12.1",
     "query-string": "5.0.1",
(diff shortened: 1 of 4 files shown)


siemiatj:
This somewhat blew out, as the performance changes broke updating field value (because widgetData is not passed directly to TableCell anymore - we're using a helper function for that - and cell wasn't aware of value changes) while fixing updating field value was breaking the original #2370 fix. Now everything seems to work fine, updated record doesn't stay visible when the page is changed, performance issues are gone and field value updates in place.

Issue #2409, opened 2019-10-11
Zoom 80% leads to awkward table-pagination layout


metas-mk:
### Is this a bug or feature request?
Bug
### What is the current behavior?
The layout is broken when zoomed out.
#### Which are the steps to reproduce?
Open Sales Order window in table view. Zoom out to 80%. The pagination moves from bottom to top right. Same happens in modal views.
![image](https://user-images.githubusercontent.com/13365687/66645930-ff6fb700-ec24-11e9-8a8e-0f9ab5b2c98e.png)
### What is the expected or desired behavior?
The pagination shall stay at the bottom again when zooming out.

Info: Behavior could have been introduced in https://github.com/metasfresh/metasfresh-webui-frontend/pull/2402

siemiatj:
More broken cases:

![Image Pasted at 2019-10-17 13-42](https://user-images.githubusercontent.com/513460/67006617-e1eb9300-f0e5-11e9-9e12-a771a2801ff9.png)


![Image Pasted at 2019-10-17 13-43](https://user-images.githubusercontent.com/513460/67006627-e57f1a00-f0e5-11e9-86c0-ed86654b4dae.png)

PR #2415, opened 2019-10-16
Fix grid's flexbox model

siemiatj:
Pagination was borked when browser zooming in/out.

Related to #2409


cristinamghita:
### Tested on local
* Hu editor is back to normal
image
* Also the layout looks ok
image

cristinamghita:
### Tested on local
* Hu editor is back to normal
image
* Also the layout looks ok
image

Issue #2401, opened 2019-10-08
Implement Google Maps Integration


TheBestPessimist:
### Is this a bug or feature request?
FR

PR #2425, opened 2019-10-25
Render results on maps

siemiatj:
Related to #2401 

https://recordit.co/58P3WnmJZS

TODO:
- [x] ~Open Street Map centering~
- [x] ~Open Street Map refreshing when changing map size doesn't work automatically~

NICE TO HAVE:
- [x] tooltip with BPartner's name when hovering over map pin
- [ ] select element on list when map marker is clicked
- [ ] select map pin + scroll map when list element is selected


siemiatj:
Tooltip when hovering over map marker

Screen Shot 2019-10-25 at 16 24 53

TheBestPessimist:
# Second functional review

> Changed the behavior of the panel style button. It know greyes-out the option that's not selected atm :

On my not-macbook-quality-display i can barely see the differece between the 2 colors (selected (=sort of blue) and disabled (=gray)). Would it make more sense to have the selected color the same blue as the "search query preview" to its left? (
image

The buttons don't feel aligned: `/` and `map` should be a little bit upper, to have their center aligned with `grid` center

siemiatj:
> # Second functional review
> 
> > Changed the behavior of the panel style button. It know greyes-out the option that's not selected atm :
> 
> #### Not enough contrast between active and disabled
> 
> On my not-macbook-quality-display i can barely see the differece between the 2 colors (selected (=sort of blue) and disabled (=gray)). Would it make more sense to have the selected color the same blue as the "search query preview" to its left?
> image

I think it'd look weird, as we don't have icons in this color anywhere. What makes more sense imho is making the grey color even lighter. And I'm testing it on a regular 1080p display too.

> #### The buttons look missaligned
> 
> `/` and `map` should be a little bit upper, to have their center aligned with `grid` center. Right now it looks as if they are "bottom aligned".

Map and Grid icons are aligned. Separator might need 1-2px padding tho.
> #### Hide the "map-change" button if OSM is the provider
> 
> Since we're not showing any maps for OSM at the moment, we shold also hide the toggle buttons since the user may press the buttons and reach a state where the maps pins should be shown, but there is an empty screen instead (like in my screenshot).
> The default view mode in this case should be "grid-only".
> image

👍

Issue #2435, opened 2019-10-31
Fix attributes dropdown closing


siemiatj:
### Is this a bug or feature request?

Bug

### What is the current behavior?

When user tries to add a new location using the `Address Dropdown`, it can't be closed and saved.

PR #2439, opened 2019-11-06
Fix patching attributes

siemiatj:
For some reason fields were still patched even when nothing has changed.

Related to #2435


metas-mk:
Will R&I later.

Issue #2436, opened 2019-10-31
Follow-Up: Customer Proposal List navigation with arrow keys (qty arrows out)


metas-mk:
### Is this a bug or feature request?
Feature Request/ Bug
### What is the current behavior?
see below
#### Which are the steps to reproduce?
Open, try and see.
### What is the expected or desired behavior?
- Behavior line selected +  focus field does not work when using action "other products" and then coming back to proposal list.
- the layout seems broken (see buttons too tight to left border)
![image](https://user-images.githubusercontent.com/13365687/67925906-f3886c80-fbb4-11e9-9ce4-c25c70d647a7.png)
- The recording keys are not set as required (arrow for navigation, enter to get into edit mode, enter to leave edit mode, arrow for further navigation). To leave edit mode one must use an arrow key.

Follow-Up of:
https://github.com/metasfresh/metasfresh-webui-frontend/pull/2432

siemiatj:
@metas-mk can you be more specific and provide more details ?

> Behavior line selected + focus field does not work when using action "other products" and then coming back to proposal list.

I don't have the `other products` action, but it seems to be working fine with `other customers`.

![nav](https://user-images.githubusercontent.com/513460/69429793-ad967280-0d34-11ea-96bc-8c5397c0be46.gif)

> The recording keys are not set as required (arrow for navigation, enter to get into edit mode, enter to leave edit mode, arrow for further navigation). To leave edit mode one must use an arrow key.

I need a description of how to reproduce that. Edit mode of what ? Where ?

PR #2464, opened 2019-11-29
Fix widgets in tables

siemiatj:
Caused by fix to #2436 , related to #2477


siemiatj:
@metas-mk am I missing something ? I'm using the dev instance from the me03 issue : https://recordit.co/dKuHQcrXo4

Issue #2472, opened 2019-12-03
Purchase order from sales order not working anymore


metas-dh:
### Is this a bug or feature request?
Bug
### What is the current behavior?
Purchase order from sales order not working anymore: when calling the action the window turns white and nothing happens.
#### Which are the steps to reproduce?
See vid in me03 task.
### What is the expected or desired behavior?
Shall work.

PR #2479, opened 2019-12-04
Fix time formatting util

siemiatj:
It was tripping over because the types didn't match.

Related to #2472


TheBestPessimist:
Dev tested locally and the modal opens now
image

PR #2539, opened 2020-01-17
Prevent unnecessary re-renders of checkbox component

petrican:


petrican:
Screenshot 2020-01-16 at 16 12 28
Prevents all those 560 re-renders to just 2

src/components/widget/Checkbox.js
```diff
@@ -42,9 +53,9 @@ const Checkbox = props => {
         <input
           ref={rawWidget}
           type="checkbox"
-          checked={widgetData[0].value}
+          checked={checkedState}
           disabled={widgetData[0].readonly || disabled}
-          onChange={e => handlePatch(widgetField, e.target.checked, id)}
+          onChange={e => updateCheckedState(e)}
```
siemiatj: why do we need an anonymous function here ?

siemiatj:
Just one thing. Other than that, LGTM

src/components/widget/Checkbox.js
petrican: This function is used to do fine tuning to state before further calling the handlePatch. In our case set the state value checkedState L:34 .

siemiatj: I still don't understand why there's an anonymous function there. If param set in onChange is just e and updateCheckedState expects e as well, why not just onChange={updateCheckedState} ?

PR #2540, opened 2020-01-20
Add unit tests for Modal component #3749

petrican:


src/__tests__/components/app/Modal.test.js
```diff
@@ -0,0 +1,410 @@
...
+import merge from 'merge';
+import { ShortcutProvider } from '../../../components/keyshortcuts/ShortcutProvider';
+
+import Modal from "../../../components/app/Modal";
+import fixtures from '../../../../test_setup/fixtures/modal.json';
+
+const mockStore = configureStore([]);
+
+
+const appHandlerState = {
```
siemiatj: Can we export states from the handlers instead ? Same goes for keymap - it's already defined in constants, just needs a wrapper function that'd accept a `mod` key, and then we could reuse it in tests too.

src/__tests__/components/app/Modal.test.js
```diff
@@ -0,0 +1,410 @@
...
+    ],
+    "ALT+ENTER": [
+
+    ],
+    "ESCAPE": [
+
+    ]
+  }
+}
+const testModal = {
```
siemiatj: This should also be in fixtures.

src/__tests__/components/app/Modal.test.js
```diff
@@ -0,0 +1,410 @@
...
+            "emptyText": "none"
+          }
+        ]
+      }
+    ],
+    "pinstanceId": "3396778",
+  }
+};
+
+const windowHandlerState = {
```
siemiatj: This can also be exported from windowHandler and used here. This way we won't have to update tests whenever we'll add anything to the state.

siemiatj (changes requested):

src/__tests__/components/app/Modal.test.js
petrican: I've updated the PR pls check. There was however a missing export of the initiaState in src/reducers/appHandler.js (fixed that as well with the changes)

src/__tests__/components/app/Modal.test.js
petrican: Fixed

siemiatj (approved):

src/__tests__/components/app/Modal.test.js
petrican: put it in there

Issue #2473, opened 2019-12-03
Performance when creating / selecting orderlines in big orders too slow


metas-dh:
### Is this a bug or feature request?
Bug
### What is the current behavior?
Performance when creating / selecting orderlines in big orders is too slow.
#### Which are the steps to reproduce?
See vid in me03 issue.

PR #2548, opened 2020-01-21
Optimizations to rendering and architecture (3rd try)

siemiatj:
Related to #2473

I'm not happy with this solution, as it does add some noise and additional code, but because some components were built this way and our data structures are like that - we can't really avoid that. Hopefully some of this will go away once we can switch tables to use redux.


siemiatj:
Marking as draft as this needs to be thoroughly tested on a dev instance, and I don't want someone to accidentally merge it.

petrican (approved):
Looks good. This should be roughly tested on the devbox by QA, e2e tests before being merged to master. Thank you

siemiatj:
Ok, got a green light from tests. Will fix the conflicts and merge.

petrican:
> Ok, got a green light from tests. Will fix the conflicts and merge.

ok.

siemiatj:
I'll merge this together with the redux state PR as I'm pretty sure there are going to be more conflicts.

petrican:
> I'll merge this together with the redux state PR as I'm pretty sure there are going to be more conflicts.

Should we wait for Monday to do this after the release ? wdyt  @siemiatj ?

siemiatj:
As mentioned before: this doesn't fix the root cause of our performance issues but cuts the lag by ~1/6th by doing two things:
- stop passing arrayslobjects via props which causes unnecessary re-renders, because this triggers a props comparison which is always false (`[] !== []`)
- switch many components to using `PureComponent` class, which does a shallow props comparison and in many cases slightly improves the performance

Before (tested by editing row and clicking the next one):
Screen Shot 2020-03-23 at 13 34 56

After:
opt1

So I'll leave this open if we want to merge this or not.

PR #2551, opened 2020-01-27
Improve speed on inputs

petrican:


src/components/widget/MasterWidget.js
```diff
@@ -30,50 +31,40 @@ class MasterWidget extends Component {
     updated: false,
     edited: false,
     data: '',
+    [this.props.fieldName]: this.props.widgetData,
```
siemiatj: Can't we just use `field` or whatever ? Not that it's ever going to change throughout the lifetime of the widget.

petrican: this.props.field is undefined ...  just  throw a console.log with it and see. I found the fieldName . to be most reliable.

petrican:
@siemiatj - can you please test again ? I refactored some of the code. Thanks

src/components/widget/MasterWidget.js
siemiatj: What I meant is just using `this.state.field` instead of using a dynamic name. I don't really see a reason for that, but also it's not that big of a deal.

PR #2552, opened 2020-01-28
Refactoring and tests around MasterWindow

siemiatj:
(very loosely at this point )Related to #2517 

This started as an effort to add `a quick fix` and tests to changes needed for #2517, but right now it's a more general thing. It includes:
- separation of `MasterWindow` to data/presentational layers using the typical react/redux component/container pattern
- moves api-specific functions away from `actions` - because they have nothing to do with redux
- adds patterns and tests for actions and reducers


petrican:
Screenshot 2020-02-04 at 13 36 05
there's seems to be an issue with a test. The weird thing on jenkins it passed. I'll re-check once it's fixed. Thank you

siemiatj:
There's nothing I can fix really. Removed my `node-modules`, did clean `npm install` and everything passes.

Screen Shot 2020-02-04 at 13 02 21

PR #2558, opened 2020-01-29
show spinner and/or message while the process is running

petrican:


src/__tests__/components/app/Loader.test.js
```diff
@@ -0,0 +1,36 @@
+import React from "react";
+import { render } from "enzyme";
```
siemiatj: Can't we shallow-render it ?

Issue #2542, opened 2020-01-21
Improve speed on textinputs 


petrican:
### Is this a bug or feature request?

B 

### What is the current behavior?
Slow text input
#### Which are the steps to reproduce?
Type...
### What is the expected or desired behavior?
Typing should be instant and fast

petrican:
<img width="1609" alt="Screenshot 2020-01-29 at 11 14 00" src="https://user-images.githubusercontent.com/1708561/73343573-7a6ae500-4289-11ea-9fbe-d7ec1cc9b0fc.png">
Go to advanced edit. -> Qty, type . You should notice speed in typing

PR #2560, opened 2020-01-30
Fix issues found in IT #2542 extra

petrican:


Jenkinsfile
```diff
@@ -34,7 +34,7 @@ So if this is a "master" build, but it was invoked by a "feature-branch" build t
 ])
 
 // set nodejs version defined in tool name of NodeJS installations located in Jenkins global plugins
-final NODEJS_TOOL_NAME="nodejs-11"
+final NODEJS_TOOL_NAME="nodejs-13"
```
siemiatj: Should we also update node in package.json and for docker images? @metas-jb  ? We're way behind the latest :)

PR #2566, opened 2020-02-01
Fix unneeded re-renderings in Table #2542

petrican:


petrican:
Finally found the issue. dataHash is sent from the BE each time and causes re-rendering all the time + the selected row is not checked

petrican:
@kuba  checked with tobi, it's fine, ignore the "red build", here is the result:  https://jenkins.metasfresh.com/job/metasfresh-dist/job/master/19451/  pls review asap. Thanks

src/components/table/TableItem.js
```diff
@@ -38,9 +39,24 @@ class TableItem extends PureComponent {
...
       multilineText,
       multilineTextLines,
+      [this.props.rowId]: this.props,
     };
   }
 
+  shouldComponentUpdate(nextProps) {
+    if (
+      !_.isEqual(
+        _.omit(nextProps, 'dataHash'),
```
siemiatj: You know that `dataHash` was deliberately added to know when data changed ?  :)

petrican: Without excluding it you will always have re-render for each or the rows. I don't think the system was built to re-render all the table rows if only one is changing/selected.

petrican:
> If it works and doesn't break anything I'm ok with this change. I'm just wondering why we're comparing the whole props if `dataHash` was added as a mean to inform us about data changes. I'm worried this will break some not-that-obvious cases, but this will come up in testing anyway.

As I see it there is something wrong in the way dataHash is sent. You would need to have dataHash sent only in correlation with the data that was modified and not with all the rows because otherwise triggers 100 re-renders for 100 rows which is insane :) . My fix does exactly this thing by differentiating what row should re-render. Thanks

Issue #2536, opened 2020-01-16
Code improvements


petrican:
### Is this a bug or feature request?

F

### What is the current behavior?

See attached

#### Which are the steps to reproduce?

npm run lint

### What is the expected or desired behavior?

Ideally no warnings

petrican:
<img width="1398" alt="Screenshot 2020-02-05 at 16 57 57" src="https://user-images.githubusercontent.com/1708561/73853024-dd83eb00-4838-11ea-8db3-534a9324d86e.png">
after the fix there should not be any warning msgs with yellow

PR #2570, opened 2020-02-03
Improve webui code #2536

petrican:


test_setup/fixtures/raw_list.json
```diff
@@ -8,14 +8,12 @@
         "key": "2000835",
         "caption": "Testpreisliste Lieferanten"
       },
-      "properties": [
-        {
+      "properties": {
```
siemiatj: weird, this didn't break anything ?

petrican: It actually fixed the the unit test which was passing an array

PR #2583, opened 2020-02-07
Fix new entry marked as red

petrican:


src/components/table/TableItem.js
```diff
@@ -46,6 +46,11 @@ class TableItem extends Component {
   }
 
   shouldComponentUpdate(nextProps, nextState) {
+    // check on saving logic
+    if (this.props.notSaved === true && nextProps.notSaved === false) {
```
petrican: Note: I could have used just this.pops.notSaved and !nextProps.notSaved but for the sake of readability and easy logic following I left it like that.

Issue #2578, opened 2020-02-06
Code quality falsely reported as degraded by codacy


petrican:
### Is this a bug or feature request?
B
### What is the current behavior?
arrow parrens need to be always present as this is how codacy is configured
#### Which are the steps to reproduce?
any code like const = something => {} would be seen as degrading
### What is the expected or desired behavior?
code should be forced by prettier on dev side to be const = (something) => {}

PR #2586, opened 2020-02-10
Change eslintrc config arrowParens #2578

petrican:


siemiatj (approved):

siemiatj:
Jesus this Codacy is such a pain in the ass. Why do we need it again ?

petrican:
> Jesus this Codacy is such a pain in the ass. Why do we need it again ?
:) 
30 % pain in the ass 70% catches some issues we might not see with local linter. I'll try to tune it on the fly as we work on issues and spot things that need to be fixed. Thanks

Issue #2591, opened 2020-02-11
mac: copy eMail address in grid view deletes the content


petrican:
### Is this a bug or feature request?
B
### What is the current behavior?
mac: copy eMail address in grid view deletes the content
#### Which are the steps to reproduce?
on a mac system
go to business partner window, select the subtab for contact
select the email address (only focus) and press cmd+c.
--> the content is deleted
### What is the expected or desired behavior?
Shall not be deleted.

PR #2592, opened 2020-02-11
Fix CMD+C combination in table cell #2591

petrican:


petrican:
Note: this was an urgent fix

Issue #2594, opened 2020-02-11
Date attributes cannot be set in HU Editor


metas-dh:
### Is this a bug or feature request?
Bug
### What is the current behavior?
When setting a date attribute, e.g. Tageslot, an error is thrown; the date is displayed at first, but not properly set for the HU.
#### Which are the steps to reproduce?
1. create a purchase order
2. open the material receipt cands, select the orderline and open the HU Editor (with e.g. "Receive HUs"): **have more than 1 HU!**
3. select a date  attribute, Tageslot: date is set, but error is thrown, NOK
4. select another HU, and check the 1st one again: date is not set anymore: NOK

see vid in me03 issue
### What is the expected or desired behavior?
Shall work.

Issue #2593, opened 2020-02-11
Screen turns white when selecting an attribute in HU Editor


metas-dh:
### Is this a bug or feature request?
Bug
### What is the current behavior?
When selecting an attribute in HU Editor, the screen turns white.
#### Which are the steps to reproduce?
1. create a purchase order
2. open the material receipt cands, select the orderline and open the HU Editor (with e.g. "Receive CUs"
3. select an attribute, e.g. Label: Bio (see also video in me03 issue)
=> screen turns white, NOK

same thin when selecting an attribute in manufacturing order, for received HU.

### What is the expected or desired behavior?
Selecting an attribute shall work.

PR #2596, opened 2020-02-11
Fix SelectionAttributes patching

siemiatj:
Related to #2593 & #2594 . Will add tests in a separate PR as there's quite a bit of preparation work needed and this can be tested in the meantime.


src/components/DataLayoutWrapper.js
```diff
@@ -122,7 +122,7 @@ class DataLayoutWrapper extends Component {
 DataLayoutWrapper.propTypes = {
   onRowEdited: PropTypes.func,
   windowType: PropTypes.string,
-  viewId: PropTypes.number,
+  viewId: PropTypes.string,
```
petrican: I noticed that the ids are either coming as a string or as a number. Normally all should be numbers right? I also put .any as that was annoying. I suspect some issue on top level components when these are passed. I guess we leave it like this and when we bump into the issue again we should fix it at the top level component sending those.

PR #2602, opened 2020-02-13
Changes to MasterWindow, tests etc

siemiatj:
Fixes error in #2517, where only the latest change was stored in redux. #2604 fixes regression caused by changes to text inputs, rewrites some of our actions to be fully async and adds some tests.

Props used for `TableItem` are just JSON.stringified props of a real component before/after changes.


siemiatj:
https://recordit.co/DzjhTQSdru

src/utils/websockets.js
```diff
@@ -34,10 +38,33 @@ export function connectWS(topic, onMessageCallback) {
...
+      brokerURL: config.WS_URL,
+      // debug: function(str) {
+      //   console.log('debug: ', str);
+      // },
+      reconnectDelay: 5000,
+      heartbeatIncoming: 4000,
+      heartbeatOutgoing: 4000,
+    });
+
+    if (process.env.JEST_WORKER_ID === undefined) {
```
petrican: quotes around undefined

src/utils/websockets.js
```diff
@@ -1,5 +1,9 @@
-import SockJs from 'sockjs-client';
-import Stomp from 'stompjs/lib/stomp.min.js';
+import SockJS from 'sockjs-client';
+import { Client } from '@stomp/stompjs';
+
+function socketFactory() {
+  return new SockJS(config.WS_URL);
```
petrican: Where is the config defined ? Here this is undefined

src/utils/websockets.js
```diff
@@ -34,10 +38,33 @@ export function connectWS(topic, onMessageCallback) {
   };
 
   const connect = () => {
-    this.sock = new SockJs(config.WS_URL);
-    this.sockClient = Stomp.Stomp.over(this.sock);
-    this.sockClient.debug = null;
-    this.sockClient.connect({}, subscribe);
+    this.sockClient = new Client({
+      brokerURL: config.WS_URL,
```
petrican: same as above comment

siemiatj: config is a global in `config.js`

src/utils/websockets.js
siemiatj: config is a global in `config.js`

src/utils/websockets.js
siemiatj: it's not `typeof` so it won't be a string

PR #2609, opened 2020-02-20
Dev 2607 multiselect

petrican:


src/components/widget/List/List.js
```diff
@@ -349,7 +349,7 @@ ListWidget.propTypes = {
   mainProperty: PropTypes.object,
   isModal: PropTypes.bool,
   autoFocus: PropTypes.bool,
-  selected: PropTypes.object,
+  selected: PropTypes.oneOfType([PropTypes.object, PropTypes.array]),
```
siemiatj: I wish we could force this to always be an array. But probably without types and with our current test coverage it'd be a nightmare :)

src/components/widget/MultiSelect.js
```diff
@@ -0,0 +1,112 @@
...
+    };
+  }
+
+  componentDidMount() {
+    this.props.onFocus();
+  }
+
+  static getDerivedStateFromProps(nextProps) {
+    let selected =
+      nextProps.selectedItems && Array.isArray(nextProps.selectedItems)
```
siemiatj: can we not have nested ternaries plx ?

src/components/widget/RawWidget.js
```diff
@@ -690,9 +697,45 @@ export class RawWidget extends Component {
             viewId={viewId}
             autoFocus={autoFocus}
             validStatus={widgetData[0].validStatus}
+            isMultiselect={isMultiselect}
           />
         );
 
+      case 'MultiListValue':
+        return (
+          <List
```
siemiatj: if the only difference is the `isMultiselect` prop, maybe we can merge this into `listWidgetProps` and use for both single and multi select lists ?

src/components/widget/MultiSelect.js
```diff
@@ -0,0 +1,112 @@
...
+      });
+      return {
+        checkedItems: updatedCheckedItems,
+      };
+    }
+
+    return null;
+  }
+
+  selectItem = (key, caption) => {
```
siemiatj: Can you add a comment on what's going on over here ? Why are we transforming `state.checkedItems` to string and then parsing it ?

src/components/widget/MultiSelect.js
```diff
@@ -0,0 +1,112 @@
...
+    let selected = null;
+    let newCheckedItems = JSON.parse(JSON.stringify(this.state.checkedItems));
+    if (typeof this.state.checkedItems[key] === 'undefined') {
```
siemiatj: Maybe we could do something like :
```javascript
_.set(checkedItems, ['key1'], { key: 'key1', value: !_.get(checkedItems, ['key1', 'value'], true)});
```

src/components/widget/MultiSelect.js
```diff
@@ -0,0 +1,112 @@
...
+    selected = Object.keys(newCheckedItems).map((item) => {
+      if (newCheckedItems[item].value === true) {
+        return {
+          key: newCheckedItems[item].key,
+          caption: newCheckedItems[item].caption,
+        };
+      }
+    });
+
+    this.props.onSelect(selected.filter((el) => typeof el !== 'undefined'));
```
siemiatj: Just wondering if we shouldn't call this after we've updated the state ? Or it won't make a difference here ?

src/components/widget/List/List.js
petrican: When is solo item selection (from dropdown) is an object.   When the selection comes from the multiselect si an array of objects

src/components/widget/RawWidget.js
petrican: This I believe is wise to be done after the BE will provide only the widget type 'MultiListValue'. For now let's leave it as it is pls.

src/components/widget/MultiSelect.js
petrican: calling it here will guarantee the top filter is set accordingly. The sync to state is done in getDerivedStateFromProps

src/components/widget/MultiSelect.js
petrican: No action should be done to directly alter the state. That is the reason I deep cloned the newCheckedItems in the first place.

petrican: Also didn't want to . add lodash and preferred the vanillaJS way :). Thanks for understanding .

src/components/widget/MultiSelect.js
petrican: Yeah that I missed. Will fix asap. Thanks

petrican: Fixed.

Issue #2514, opened 2020-01-02
Tree dots click triggers console error/warning


petrican:
### Is this a bug or feature request?

Bug

### What is the current behavior?

Being authenticated when you click on the three dots in the console a warning/error is shown. No major impact on UI rendering.

#### Which are the steps to reproduce?

Clicking on the three dots triggers an error/warning in the console

### What is the expected or desired behavior?

There should be no error in console and the case flagged by the error should be considered

siemiatj:
This looks like a good `easy win` so feel free to grab it whenever you'll get blocked on other things or in between the tasks.

petrican:
<img width="923" alt="Screenshot 2020-02-26 at 16 48 51" src="https://user-images.githubusercontent.com/1708561/75355942-ead04a80-58b7-11ea-80e3-aeae7fc8ce4b.png">
DOC

PR #2620, opened 2020-02-26
fix situation where window is not defined #2514

petrican:
There are situations where window is not defined like in the root of the site


petrican:
https://chat.metasfresh.org/metasfresh/pl/end6zq3u578p9q1eiruu5bgmio

Issue #2632, opened 2020-03-04
Have the languages in store


petrican:
### Is this a bug or feature request?
B
### What is the current behavior?
Cypress tests look for those in there and they are not found
#### Which are the steps to reproduce?
Run cypress test flatrate_conditions_material_tracking and it will crash on country
### What is the expected or desired behavior?
Tests should pass after this addition

PR #2633, opened 2020-03-04
Add languages into the redux store #2632

petrican:
Adds the languages into the redux store. This is needed because the existing cypress tests were looking into these and they could not be found..thus tests failing.

### DIFF
--- a/src/actions/AppActions.js
+++ b/src/actions/AppActions.js
@@ -374,6 +374,13 @@ export function userSessionUpdate(me) {
   };
 }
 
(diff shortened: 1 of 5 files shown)


petrican:
DO not merge yet! Even if the values (languages) are needed in the store. Updating them seems to cause a loop.

petrican:
> DO not merge yet! Even if the values (languages) are needed in the store. Updating them seems to cause a loop.

Pushed fix should be fine now

Issue #2613, opened 2020-02-25
Right click + edit field value doesn't work 


metas-dh:
### Is this a bug or feature request?
Bug
### What is the current behavior?
Right click + edit field value doesn't work 
#### Which are the steps to reproduce?
Try and see: e.g. in window product prices, try with the price field.
### What is the expected or desired behavior?
Should work.

PR #2636, opened 2020-03-05
Fix right click context menu

siemiatj:
Probably broken due to upgrading react/redux recently, which changed how refs are passed. Small refactoring along the way.

Related to #2613 .

### DIFF
--- a/src/components/table/Table.js
+++ b/src/components/table/Table.js
@@ -951,13 +951,17 @@ class Table extends Component {
     }
   };
 
-  handleFieldEdit = (selected, fieldName) => {
-    this.closeContextMenu();
+  handleFieldEdit = () => {
+    const { selected, contextMenu } = this.state;
 
-    const selectedId = selected[0];
+    if (contextMenu.supportFieldEdit && selected.length === 1) {
+      const selectedId = selected[0];
 
-    if (this.rowRefs && this.rowRefs[selectedId]) {
-      this.rowRefs[selectedId].initPropertyEditor(fieldName);
+      this.closeContextMenu();
+
+      if (this.rowRefs && this.rowRefs[selectedId]) {
+        this.rowRefs[selectedId].initPropertyEditor(contextMenu.fieldName);
+      }
     }
   };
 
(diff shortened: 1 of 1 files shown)


src/components/table/Table.js
```diff
@@ -951,13 +953,17 @@ class Table extends Component {
     }
   };
 
-  handleFieldEdit = (selected, fieldName) => {
-    this.closeContextMenu();
+  handleFieldEdit = () => {
+    const { selected, contextMenu } = this.state;
 
-    const selectedId = selected[0];
+    if (contextMenu.supportFieldEdit && selected.length === 1) {
```
petrican: I believe selected can be undefined as well. Case in which this will trigger an error. Can you pls add a safety check for that as well or define it in the state as [] in constructor ? Thanks

src/components/table/Table.js
```diff
@@ -1060,7 +1066,8 @@ class Table extends Component {
         ref={(c) => {
           if (c) {
             const keyProp = item[keyProperty];
-            this.rowRefs[keyProp] = c.wrappedInstance;
+
+            this.rowRefs[keyProp] = c;
```
petrican: Can you pls refactor a bit and put a meaningful name instead of just c? Thank you

petrican: Also was wondering what do you think about the check from codacy -> https://app.codacy.com/gh/metasfresh/metasfresh-webui-frontend/pullRequest?prid=5129401. Should we remove that ? we could forcely set the parseInt for some to fix it but code would look ugly..  I believe we could ditch that  Generic Object Injection Sink check. Let me know your thoughts

src/components/table/Table.js
siemiatj: selected[0] can be undefined. Otherwise as far as I'm aware it should always be an array. It is defined as one in the state in `tableHelpers`.

src/components/table/Table.js
siemiatj: Yeah, let's remove that. It's a dumb and completely unnecessary check.

siemiatj: Do we really need a meaningful name in `ref` function ? We all know how it works, and changing this in one place would mean we have to change it everywhere to stay consistent. `c` or `ref` is what we're using all across the app.

petrican: Would be nice to have it -> https://github.com/ryanmcdermott/clean-code-javascript#variables (Avoid Mental Mapping)

petrican: ^^ only referring to (c)

siemiatj: Yeah but then again we shouldn't get crazy about it. `c` in ref is always a component.

PR #2653, opened 2020-03-24
Revert "Hotfix to keep value in the state and update it with a debounced call…"

petrican:
Reverts metasfresh/metasfresh-webui-frontend#2650


petrican:
Reverting this as it breaks other another functionality

Issue #2652, opened 2020-03-24
Remove unnecessary scrollbars


siemiatj:
### Is this a bug or feature request?

Bug

### What is the current behavior?

There are excessive scrollbars round the windows and in some fields on Chrome/Windows.

PR #2654, opened 2020-03-24
Fix window scrollbars on Windows

siemiatj:
Related to #2652

### DIFF
--- a/src/assets/css/panel.css
+++ b/src/assets/css/panel.css
@@ -94,7 +94,8 @@
 }
 
 .panel-bordered-force {
-    overflow: auto;
+    overflow-x: auto;
+    overflow-y: hidden;
 }
 
 .panel-primary {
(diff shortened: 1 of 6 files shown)


src/assets/css/panel.css
```diff
@@ -94,7 +94,8 @@
 }
 
 .panel-bordered-force {
-    overflow: auto;
+    overflow-x: auto;
+    overflow-y: hidden;
```
siemiatj: I can't really see any difference in HU editor but it doesn't break anything either so maybe I'm missing something. Thanks

TheBestPessimist: The issue happened only on some specific cases, ie. on page 1 there's no scrollbar, but on page 2 there is; then on page 3 there isnt one again. I think it's related to the content of the pages.

Issue #2668, opened 2020-04-03
`Materialentnahme`=`Material Withdrawal` refresh issues


petrican:
### Is this a bug or feature request?
B
### What is the current behavior?
see m03 issue 
#### Which are the steps to reproduce?
https://github.com/metasfresh/me03/issues/4346

PR #2669, opened 2020-04-03
Setting correct view on filter change #2668

petrican:


src/components/app/DocumentList/index.js
```diff
@@ -424,7 +420,7 @@ class DocumentListContainer extends Component {
     filterView(windowType, viewId, filtersActive.toIndexedSeq().toArray())
       .then((response) => {
         const viewId = response.viewId;
-
+        this.connectWebSocket(viewId);
```
siemiatj, line 423: Do we have to do this for all filters or barcode scanner is just some edge case ?

siemiatj:
Code-wise it's ok, but want to be sure about two things:
- are we unsubscribing the previous websocket when filtering ?
- do we have to re-subscribe for every filtering or just barcode scanner ?

src/components/app/DocumentList/index.js
petrican: For all filters. If you look at the URI when you setup new filter the viewId is changing

TheBestPessimist:
@siemiatj From my testing, it seems that each change in a filter will return a new view/viewID. 
I'm not sure if this needs the unsubscribe old -> subscribe new, but it is my understanding (from testing on local machine) that it should be like that, as since backend changed the view, the old one has stale data.

PR #2673, opened 2020-04-07
Dev 2668 material witdrawal 2

petrican:


TheBestPessimist:
Results of Dev test here: https://github.com/metasfresh/me03/issues/4346#issuecomment-610367704

PR #2677, opened 2020-04-09
Add redux skeleton for filters and a flag to know when all are to be …

petrican:
- added skeleton for having filters data in redux
- added a flag in the store for knowing when filters are to be cleared
- adapted the code to use this flag
Note: The entire filter logic should reside in the redux store instead of using state at component logic. That implies refactoring and adding separate actions and dispatch them instead of setting the state within component. This I would like to discuss in our next frontend meeting


src/components/header/Breadcrumb.js
```diff
@@ -284,6 +287,15 @@ Breadcrumb.propTypes = {
   openModal: PropTypes.any,
   siteName: PropTypes.any,
   windowType: PropTypes.any,
+  filters: PropTypes.object,
+};
+
+const mapStateToProps = (state) => {
```
siemiatj, line 293: Why not just simplify this with 

```
mapStateToProps = ( { filters }) => { filters };
```

src/components/header/Breadcrumb.js
```diff
@@ -97,7 +97,10 @@ class Breadcrumb extends Component {
    * @param {*} menu
    */
   handleClick = (e, menu) => {
-    const { handleMenuOverlay, windowType } = this.props;
+    const { handleMenuOverlay, windowType, filters, dispatch } = this.props;
+    if (!filters.clearAll) {
+      dispatch(clearAllFilters(true));
```
siemiatj, line 102: maybe instead of passing `dispatch` just throw `clearAllFilters` in the connect, like we have elsewhere in the app ?

src/constants/FilterTypes.js
```diff
@@ -0,0 +1,3 @@
+// FILTTER ACTIONS
```
siemiatj, line 1: Can you rename the file to `FilterActionTypes` instead ?

siemiatj:
Just minor thing and some suggestions. Other than that, LGTM

src/constants/FilterTypes.js
petrican: All files under constants are ItemTypes, ListTypes I put the name FilterTypes in order to follow the pattern. If I were to change it it would break the pattern.

src/components/header/Breadcrumb.js
petrican: You mean add it within mapDispatchToProps.  You see in this file everything is called with dispatch so I would leave this for the next refactoring as this is an urgent fix that needs to reach master.  Basically every action would need to reside in mapDispatchToProps. Thanks for the suggestion

Issue #2676, opened 2020-04-09
Spinner over modal when opened from grid


siemiatj:
### Is this a bug or feature request?

Bug

### What is the current behavior?

When running `open linked accounting transactions` action from grid view, the spinner doesn't disappear (and is misaligned).

#### Which are the steps to reproduce?

Go to `window/162` (or `accounting transactions`), select a row and run `Open linked accounting transactions` action.

### What is the expected or desired behavior?

Spinner should hide.

PR #2678, opened 2020-04-09
Fix handling data in modals from grid view

siemiatj:
Related to #2676

### DIFF
--- a/src/components/app/DocumentList/index.js
+++ b/src/components/app/DocumentList/index.js
@@ -203,7 +203,7 @@ class DocumentListContainer extends Component {
               deselectTableItems(removedRows, windowType, viewId);
             } else {
               if (filtersActive.size) {
-                this.filterView();
+                this.filterCurrentView();
               }
 
               // force updating actions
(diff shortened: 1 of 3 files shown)


src/components/app/DocumentList/index.js
```diff
@@ -300,9 +304,10 @@ class DocumentListContainer extends Component {
       fetchLayout,
       updateRawModal,
       setModalDescription,
+      isModal,
     } = this.props;
 
-    fetchLayout(windowType, type, viewProfileId)
+    fetchLayout(windowType, type, viewProfileId, isModal ? viewId : null)
```
petrican, line 310: Can we not have such logic and pass only a param name ?  There are way too many params anyway in here I understand that that cannot be changed now and pass only one single object but let's just pass  a variable in here with a meaningful name in here and it's fine.

petrican:
Merging it as it is as I understand from Ruxi that it has to reach erpit today

src/components/app/DocumentList/index.js
siemiatj: Yeah, I know. Will refactor this as I might be working near this code now.

PR #2688, opened 2020-04-15
Add Table redux skeleton and add flag for knowing sort actions happen

petrican:
- fix for sorting
- added skel for future redux data for the table
- allow re-rendering when flag is found in the redux store


siemiatj:
I'm not convinced if we should have a new `table` reducer. Tables are either part of the grid view or details view, where we now store the rest of the data (or should store, as filters are mostly stored locally). It's gonna get problematic when we'll have a details view with modal opened, and we'll try to get sorting for both of them at once. Table reducer will have to search by `windowType /documentId` for the details view, and by `windowType/viewId` for the modal simultaneously.

Unless this is some other sorting.

petrican:
As discussed on Skype we need to have the logic separately and not to have dup code from the grid and details view. We will merge this as it is and with the next refactoring of the table area we will clean up even more. Thanks

Issue #2693, opened 2020-04-17
Fix selections in grids


siemiatj:
### Is this a bug or feature request?

Bug

### What is the current behavior?

When loading grid view, first row is sometimes selected, and other times not. Sometimes selection is properly stored in the redux store but not indicated in the list.

PR #2695, opened 2020-04-21
Cleanup selections logic

siemiatj:
Trying to make some sense of selections in Tables. Also added a small helper for getting quickActions.

Related to #2693

### DIFF
--- a/src/api/view.js
+++ b/src/api/view.js
@@ -205,29 +205,33 @@ export async function quickActionsRequest(
   parentView
 ) {
   const requests = [];
+
   const query = getQueryString({
     viewProfileId,
     selectedIds,
-    childViewId: childView.viewId,
-    childViewSelectedIds: childView.viewSelectedIds,
-    parentViewId: parentView.viewId,
-    parentViewSelectedIds: parentView.viewSelectedIds,
   });
+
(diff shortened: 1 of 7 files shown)


petrican:
Pls try to fix the selections when you switch back from right to left panels and make sure the selection is made by default when you select on the right with either first if no previous selection is done or last from the left side if there was an already made selection. https://github.com/metasfresh/me03/issues/4490#issuecomment-617593564. 

Would have been nice to have this in a table reducer as we are currently planning but I don't think there is enough time for such refactoring now.

Like:
table {
    active_selections: {
          [viewId_tabID] : rowId
    } 
}

This way you would know anytime what is selected and where.

siemiatj:
So the related e2e tests seem to pass now : https://dashboard.cypress.io/projects/5yp4q1/runs/745/specs . Some other stuff failed, looks to be mostly a timing issue or something.

petrican:
Yeah picking/pick_HUs_and_create_shipment.js was the one failing as a result of bad selection.
The others tests are failing due to something else. Dunno. Might be what Tobi said he merged. Anyways will check those starting next week

petrican:
Lgtm. Cypress suite that used to work is passing now. Good job!

PR #2699, opened 2020-04-24
Initial skeleton to render the modal when menu item is accessed #6504

petrican:
Remaining: 
- uncomment the code that checks the standardActions when BE adds comments to the standard actions
- add unit tests


src/components/header/SubHeader.js
```diff
@@ -263,7 +277,7 @@ class SubHeader extends Component {
       handleLetter,
       handlePrint,
       openModal,
-      standardActions,
+      // standardActions,   // Todo: !!! add back this for filtering the standard actions
```
petrican, line 280: comments is not yet in the standard actions . That is the reason I commented this. Will add back once it is present in the BE response.

src/components/header/SubHeader.js
```diff
@@ -326,8 +340,17 @@ class SubHeader extends Component {
...
+        handler: () => {
+          this.handleComments();
+        },
+        icon: 'meta-icon-message',
+        caption: counterpart.translate('window.comments.caption'),
+        hotkey: keymap.OPEN_NOTES,
+      },
     ]
-      .filter((docLink) => standardActions.has(docLink.action))
+      // .filter((docLink) => standardActions.has(docLink.action))  // uncomment this and add to standard actions the notes
```
petrican, line 353: Same situation as above. Bypassed the check.

src/actions/WindowActions.js
```diff
@@ -761,6 +766,113 @@ export function fetchTopActions(windowType, docId, tabId) {
   };
 }
 
+const getAPIUrl = function({ windowId, docId, tabId, rowId, path }) {
```
siemiatj, line 769: If it's just an API call, why not move it to `/api` ?

src/actions/WindowActions.js
```diff
@@ -761,6 +766,113 @@ export function fetchTopActions(windowType, docId, tabId) {
...
+  if (!docId && rowId) {
+    documentId = rowId[0];
+  }
+
+  return `${config.API_URL}/window/${windowId}${
+    documentId ? `/${documentId}` : ''
+  }${rowId && tabId ? `/${tabId}/${rowId}` : ''}/${path}`;
+};
+
+export function callAPI({ windowId, docId, tabId, rowId, target, verb, data }) {
```
siemiatj, line 781: I'd give it a more meaningful name related to what it actually does, where and what for.

src/actions/WindowActions.js
```diff
@@ -761,6 +766,113 @@ export function fetchTopActions(windowType, docId, tabId) {
...
+        return data;
+      });
+    }
+  };
+}
+
+/**
+ * Formats the url for the api call
+ */
+export function formatParentUrl({ windowId, docId, rowId, target }) {
```
siemiatj, line 844: Will we use this in cases other than comments ?

src/actions/WindowActions.js
```diff
@@ -761,6 +766,113 @@ export function fetchTopActions(windowType, docId, tabId) {
...
+  }
+  return parentUrl;
+}
+
+/**
+ * This can be further adapted to allow pre-formatting of the data before post
+ * @param {string} target
+ * @param {string} data
+ */
+export function preFormatPostDATA({ target, postData }) {
```
siemiatj, line 868: This should land in utils.

src/components/comments/CommentsPanelForm.js
```diff
@@ -0,0 +1,93 @@
...
+
+  /**
+   * Focus by default on the text input
+   */
+  useEffect(() => {
+    textRef.current.focus();
+  }, []);
+
+  return (
+    <div className="col-12">
```
siemiatj, line 21: If this is finally a valid form, why not use the `form` element ?

src/components/comments/CommentsPanelForm.js
```diff
@@ -0,0 +1,93 @@
...
+  windowId: PropTypes.string,
+};
+
+const mapStateToProps = ({ commentsPanel }) => ({
+  textInput: commentsPanel.textInput,
+});
+
+const mapDispatchToProps = (dispatch) => {
+  return {
+    updateText: (data) => dispatch(updateCommentsPanelTextInput(data)),
```
siemiatj, line 85: I don't really see the reason to create those function just to change their names. Won't using shorthand notation work for functional component or something ?

src/components/comments/CommentsPanelListing.js
```diff
@@ -0,0 +1,31 @@
...
+  return {
+    comments: data,
+  };
+};
+
+CommentsPanelListing.propTypes = {
+  comments: PropTypes.array,
+};
+
+export default connect(mapStateToProps)(CommentsPanelListing);
```
siemiatj, line 31: Just thinking out loud - wouldn't connecting just the parent comments panel be more efficient ? One `connect` that fetches data and decorates the action creators instead of two ?

src/components/header/SubHeader.js
```diff
@@ -326,6 +327,15 @@ class SubHeader extends Component {
         caption: counterpart.translate('window.Delete.caption'),
         hotkey: keymap.DELETE_DOCUMENT,
       },
+      {
+        action: 'comments',
+        handler: () => {
```
siemiatj, line 332: Can't we just use `handler: handleComments` instead of creating an anonymous function ? I just noticed we're doing the same for other actions but I don't really see a reason (unless I'm missing something obvious).

src/constants/CommentTypes.js
```diff
@@ -0,0 +1,7 @@
+// COMMENTS ENTITY ACTIONS
```
siemiatj, line 1: I'm thinking how we can better approach this as the `ActionTypes` file has started growing really fast. Maybe `ActionTypes.js` should import from other files and then export everything ?

siemiatj:
One functional thing - shouldn't we update comments list after submitting if we're already getting the updated data in the response ?

src/components/comments/CommentsPanelForm.js
petrican: changed

src/constants/CommentTypes.js
petrican: There should be an index.js in the root that exports each individual action types(CommentTypes, FIlterTypes aso). This way you have the option to import in other files either all actions types or the ones specific for the desired goal. We should not aggregate all in ActionTypes. Pls consider this in the refactoring. Actually I am going to put this comment in there such that we won't forget. Thanks

src/components/header/SubHeader.js
petrican: Yeah you are right. Changed it. Thanks

src/actions/WindowActions.js
petrican: yes it can be reused. Will put it somewhere under the api/

src/actions/WindowActions.js
petrican: Moved

src/actions/WindowActions.js
petrican: Moved this also

src/components/comments/CommentsPanelForm.js
petrican: You're right I updated it. Thanks

src/components/comments/CommentsPanelListing.js
petrican: It won't in this case. You have to think also at renderings involved.  You see CommentsPanel re-renders only when windowId or docId change. If I were to do that would trigger re-renderings when props change(from connect).  I only want to re-render the CommentsPanelListing when the data within it changes.

src/actions/WindowActions.js
petrican: It's actually a meaningful name for what it does. I will just add a proper description for the function as I missed that

PR #2702, opened 2020-04-27
Fix closing RawModal

siemiatj:
Regression from #2695 . Related to #2693


petrican:
Since we reverted the other PR will close this one. Thanks

PR #2705, opened 2020-05-05
Redux plumbing for tables

siemiatj:
I'm thinking we should maybe also move table-related stuff from the layout to the table tree ?

### DIFF
--- a/src/constants/ActionTypes.js
+++ b/src/constants/ActionTypes.js
@@ -1,3 +1,5 @@
+export * from './TableTypes';
+
 /**
  * @constant
  * @type {string} ToDo: Description for the constant.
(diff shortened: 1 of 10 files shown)


src/constants/ActionTypes.js
```diff
@@ -1,3 +1,5 @@
+export * from './TableTypes';
```
petrican, line 1: Shouldn't these be in the root of the src/constants/ in an index.js file ?

src/reducers/index.js
```diff
@@ -9,6 +9,7 @@ import viewHandler from './viewHandler';
 import filters from './filterHandler';
 import table from './tableHandler';
 import commentsPanel from './commentsPanel';
+import tables from './tables';
```
petrican, line 12: can you pls with the refactoring move the logic from the tableHandler (L: 10 above to move to  tables) ?  Or should we use 'table' for the entities ?!? The idea is that we should agree on a form. Either plural or singular and then have consistency across the app.

petrican:
> I'm thinking we should maybe also move table-related stuff from the layout to the table tree ?

Yes. That's the idea to have a reusable component independent of the layout where it renders

